PR #10, opened 2018-08-24
use average species omega in collisions

sjplimp:
## Purpose

Some reaction bug fixes, in particular to use the omega value averaged over species
in the reaction.

## Author(s)

Zhi-Hui Wang (University of Chinese Academy of Sciences (UCAS))
and Israel Borges Sebastiao (Purdue University) reported these bugs
and suggested changes.

## Backward Compatibility

Could change outputs for collisional flows.

## Implementation Notes

## Post Submission Checklist

### DIFF
--- a/src/react_bird.cpp
+++ b/src/react_bird.cpp
@@ -228,11 +228,12 @@ void ReactBird::init()
     if (r->type == IONIZATION) z = 0.0;
     
     // add additional coeff for effective DOF
+    // added MAX() limit, 24Aug18
 
     double c1 = MY_PIS*epsilon*r->coeff[2]/(2.0*sigma) *
       sqrt(mr/(2.0*update->boltz*tref)) *
       pow(tref,1.0-omega)/pow(update->boltz,r->coeff[3]-1.0+omega) *
-      tgamma(z+2.5-omega)/tgamma(z+r->coeff[3]+1.5);
+      tgamma(z+2.5-omega) / MAX(1.0e-6,tgamma(z+r->coeff[3]+1.5));
     double c2 = r->coeff[3] - 1.0 + omega;
 
     r->coeff[2] = c1;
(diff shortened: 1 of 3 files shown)


sjplimp:
@stanmoore1 ditto on this one, received files from Michael

aborner1:
@sjplimp Should the rotrel and vibrel functions also be modified to use the average species omega?

sjplimp:
this is a good Q for Michael - I will ask him.

src/react_bird.cpp
```diff
@@ -228,11 +228,12 @@ void ReactBird::init()
     if (r->type == IONIZATION) z = 0.0;
     
     // add additional coeff for effective DOF
+    // added MAX() limit, 24Aug18
 
     double c1 = MY_PIS*epsilon*r->coeff[2]/(2.0*sigma) *
       sqrt(mr/(2.0*update->boltz*tref)) *
       pow(tref,1.0-omega)/pow(update->boltz,r->coeff[3]-1.0+omega) *
-      tgamma(z+2.5-omega)/tgamma(z+r->coeff[3]+1.5);
+      tgamma(z+2.5-omega) / MAX(1.0e-6,tgamma(z+r->coeff[3]+1.5));
```
zhangchonglin: I am testing a simple flow over 2D cylinder case with 11 species (including both neutral and ionization reactions with ambipolar approximation). 

When I use any version after this commit, at around time step 6544, the number of test collisions will suddenly increase by 10-50 times, and the number of actual collisions will suddenly increase by at least 10 to 500 times, after that time step. This is happening when the total number of particles in the simulation does not change too much. 

I attached my test case, and 2 log files corresponding to results from 2 commits:
(1) log file: sim_log.2b038d8 corresponds to commit 2b038d810b3a285a9d7dc53519d417a8ec9f69bf, this is after the above modification. At step 6544, the number of test collisions and the number of actual collisions will suddenly increase. The flow field result seems wrong in this case.
(2) log file: sim_log.45e17cd corresponds to commit 45e17cdbda3845e40ed5aaf44399323c365b2941, just before the above modification. The number of test and collisions in each time step stays reasonable, relative to the total number of particles present. 

Could you please take a look and see if there is any problem introduced with this modification.  
[2d_cylinder_11sp.zip](https://github.com/sparta/sparta/files/2704655/2d_cylinder_11sp.zip)

zhangchonglin: It seems to be due to the modification of the above line of code: 
from:      tgamma(z+2.5-omega)/tgamma(z+r->coeff[3]+1.5);
to:      tgamma(z+2.5-omega) / MAX(1.0e-6,tgamma(z+r->coeff[3]+1.5));

PR #24, opened 2018-12-10
Fix overflow issue with large grids

stanmoore1:
## Purpose

Fix overflow issue with large grids

## Author(s)

Stan Moore (Sandia)

## Backward Compatibility

No issues.

I tested this for a large 2D problem where `Nx_grid * Ny_grid` was larger than 2 billion.


stanmoore1:
I knew you were going to ask that. I'm not 100% sure, but it crashes if they aren't.

stanmoore1:
I tested this PR using a very large 2D grid (over 2 billion elements) and it fixes the overflow error.

PR #28, opened 2018-12-13
Interspecies coll params

espiekos:
## Purpose

Add an ability to specify collision parameters for an interspecies pair, rather than just using the average for the participating species.

Reduced mass was also added to the params array so we didn't have to constantly recalculate it.

## Author(s)

Ed Piekos

## Backward Compatibility

VSS input file changes are fully backward-compatible.

## Implementation Notes

Michael has tested and approved the changes.

## Post Submission Checklist

- [ x] The feature or features in this pull request is complete
- [ x] Suitable new documentation files and/or updates to the existing docs are included
- [ x] One or more example input decks are included
- [ x] The source code follows the SPARTA formatting guidelines


stanmoore1:
@espiekos Sorry for the delay. The `doc/collide.html` file is auto-generated from the corresponding `doc/collide.txt file`, could you update the text file as well?

stanmoore1:
All regression tests pass except `/examples/ambi` which has numerical differences.

stanmoore1:
@espiekos are the differences in `/examples/ambi` an expected change?

espiekos:
Interestingly, at least with respect to the trace species particle counts in the log file, most of the diff comes from a previous checkin unrelated to this one.  

If I run the code currently on the download site, c10[6-11] at step 1000 are:
25            2           94           747           19            0

where, after this change, they are:
38            8          110          795           27            0

The single processor log file in the example, however, has:
112           6          334          455           28            0

One change I made that does affect the results somewhat is to use the new average omega in a few places where a single-species omega was being used (lines 449 and 483 in the modified collide_vss.cpp).  I have talked to Michael about this issue and he says that the new omega is correct.  If I revert these lines to the old behavior, the counts match the 4Jan19 release results exactly.

stanmoore1:
@espiekos FYI we want to test for performance regression of this change before merging to make sure the 2D param array doesn't cause a noticable slowdown.

stanmoore1:
All the regression tests passed for this PR.

stanmoore1:
These changes were approved by Michael.

PR #29, opened 2018-12-18
Reaction and relaxation fixes

### LINKED ISSUE #2: Variable Rotational Relaxation
## Summary

Current variable rotational relaxation approach does not reproduce the expected relaxation rate.

## Type of Issue

Physics Bug

## Detailed Description

The current version of the rotational relaxation model does not correctly reproduce the expected relaxation rate and will not reach equilibrium. In the current model, the energy of the colliding particles is used in Parker's model to calculate Z_rot, `rotphi` in the code. Using `Ec` to calculate Z_rot requires a correction factor that Boyd developed in the past. This is quickly summarized in Boyd and Schwartzentruber, _Nonequilibrium Gas Dynamics and Molecular Simulation_, pp. 254-5. 

I typically prefer using the cell temperature approach where `Tr` is replaced with the cell temperature. I suggest using the cell translational temperature instead of calculating a composite cell temperature. This is also the approach taken by Boyd and Schwartzentruber, pp. 253-4.   

The vibrational relaxation number, Z_vib or `vibphi` in the code, may also suffer from the same issue. Boyd and Schwartzentruber, pp. 260-5 discusses vibrational relaxation. In my code, I assume the translational temperature for the Millikan-White model.

## SPARTA Version

`collide_vss.cpp`

Rotation:
https://github.com/sparta/sparta/blob/45e17cdbda3845e40ed5aaf44399323c365b2941/src/collide_vss.cpp#L773

Vibration:
https://github.com/sparta/sparta/blob/45e17cdbda3845e40ed5aaf44399323c365b2941/src/collide_vss.cpp#L785

## Expected Behavior

Thermal equilibrium should be obtained in a relaxation case.

## Actual Behavior

When initializing a simple heat bath simulation at equilibrium, the temperatures diverge to some steady nonequilibrium state when using variable rotational relaxation. The correct result is observed when using a constant relaxation number.

## Steps to Reproduce

Run a heat bath simulation with N2 where all temperatures start at an equilibrium of ~10,000 K.

aborner1:
## Purpose

Adds ability to pick between 2 particle selection methodologies for non-reactive inelastic collisions: serial implementation from Bird and particle selection prohibiting double relaxation from Haas et al.
Adds corrections for rotational and vibrational relaxation numbers to match values from continuum.
Adds new implementations to variable relaxation numbers.
Corrects some signs in post-collision velocity calculations for scattering subroutines.

Adds ability to pick between Bird's DS1V implementation for energy used in TCE model, and total energy (sum of trans, rot and vib).
Implements using the 3rd body number density rather than total number density for recombination reactions.
Fixes a bug where particles where not deleted properly for recombination reactions specified with blank/mol/atom keywords.

Fixes reading the participating internal DOF directly from .tce file.

Fixes #2 when using the particle selection methodology prohibiting double relaxation and variable rotational relaxation number.

## Author(s)

Arnaud Borner (NASA Ames Research Center)
Israel Borges Sebastiao (Pfizer/Purdue University)

## Backward Compatibility

New rotational and vibrational relaxation numbers (constant and variable) break backward compatibility.

## Implementation Notes

Changes approved by Michael prior to push.

## Post Submission Checklist


stanmoore1:
@aborner1 I'm seeing numerical differences even for the simple collide benchmark, is that expected?

aborner1:
@stanmoore1 There were some bug fixes included in this large release (on top of my head I can think of at least some signs in the post collision velocity computations), so I would likely expect numerical differences even in a simple case of a monoatomic colliding at equilibrium. I could confirm this by testing the master version with the bug fixes (without all the new added features) and the original master.

stanmoore1:
@aborner1 
>I could confirm this by testing the master version with the bug fixes (without all the new added features) and the original master.

That would be helpful. I think the main issue why this hasn't been merged yet is that it is hard to ensure we aren't introducing a new bug with all of the changes lumped together. @mgallis asked me to fix the velocity sign issue, see #59, so I put that in as a separate PR.

stanmoore1:
@aborner1 actually the bug fixes got merged in by #37.

stanmoore1:
@aborner1 after merging in #37, I'm still seeing a few regression tests failing due to this PR. Specifically these:

```
examples/axi
examples/vibrate
examples/ambi
examples/chem
```

Any ideas what could have changed?

aborner1:
@stanmoore1 Let me look into this right now

aborner1:
@stanmoore1 I just successfully ran all 4 examples you mentioned both serially and on 16 ranks with current master c42802f. Do you have more info on what makes the regression testing fail?

stanmoore1:
@aborner1 I'm comparing the stats output in the logfiles, and there were numerical differences in those tests, so the trajectories changed. I used the `global comm/sort yes` flag to ensure that the trajectories are the same between runs (when using the same # of MPI ranks).

stanmoore1:
And to clarify, the regression tests only fail with this PR (#29) merged in, comparing to latest master as the "correct" answer.

stanmoore1:
So something in #29 is changing the numerics of those 4 tests. Either a bugfix is included in #29 that hasn't been merged in to master yet, or a new bug was introduced in #29. We need to figure out which it is before we can merge #29.

aborner1:
@stanmoore1 Apologies, I now realized you merge master into react-fixes, I initially thought it was the opposite. So it makes sense.
The 4 cases that lead to different results are the ones that use polyatomic gases (N2, CO2 or both), with or without chemistry.
The air.species and co2.species files in each test folders have to be replaced by the ones I modified in /data/. Now, instead of reading the relaxation probability directly, we read the relaxation number (which is what it is called in the header of the *.species file). So the results with new code and old input files will be wrong.

aborner1:
@stanmoore1 Do you want me to commit the new files directly in each folder?

aborner1:
@stanmoore1 I have modified the files I believe are necessary. Please try again and let me know if any regression testing fails, and if so, which cases do. Thanks

stanmoore1:
Ah, you modified the species and vib files, which needed to be propagated to the example folders, right?

aborner1:
@stanmoore1 Correct, and the vss file as well, but only the "ambi" case used variable relaxation parameters, so it only affected that case. Finally, I changed the default TCE model to "TCE", but the old one was "Bird", so to reproduce the old results, I had to specify "react_modify bird" in the input script.

stanmoore1:
@aborner1 I'm running your changes through regression testing again

stanmoore1:
@aborner1 can this PR be closed, or are there changes that still need to be merged in?

aborner1:
@stanmoore1 Pretty much nothing got merged in except for a few select changes

stanmoore1:
OK, I forget why this got hung up, I think I'll send an email offline.

aborner1:
Closing this, some of these features are reappearing in newer PRs, but they will be separate for better tracking and understanding.

PR #41, opened 2019-02-12
Implicit

sjplimp:
## Purpose

Add support for 2 kinds of surface elements: explict (same as before) and implicit.
The latter are defined by grid corner point values read from a file.  Also enable
explict surfs to be distributed across procs so that models with huge surf element
counts can be modeled.  Implicit surfs are always distributed.

## Author(s)

Steve

## Backward Compatibility

N/A.

## Implementation Notes

## Post Submission Checklist


sjplimp:
@stanmoore1 I still want to add an example or two, and a Python tool, but otherwise I think this is ready for you to look at and regress test.  It passed all my testing.  This is not fully funcational for distributed
or implicit surfs, in the sense of all SPARTA commands support them,  But I think there are error checks with messages for the commands that don't.

stanmoore1:
@sjplimp sounds good, I'll run this through regression testing.

sjplimp:
@stanmoore1 Also there is no checking in the Kokkos package for these kinds
of new surfs, so that should be added.

sjplimp:
@stanmoore1 Added the Python surf generation tools and several example scripts and dirs.
Note that in examples/jagged there is a 3d sdata file that is too big to include in the distro.
See the README for how to generate it via a provided Python script.

sjplimp:
@stanmoore1 fixed the compute reduce but with the new surf data structs

PR #42, opened 2019-02-13
adapt_grid help doc bug

zhangh3:
## Purpose

fix  adapt_grid help doc bugs


stanmoore1:
@zhangh3 good catch, thanks for pointing this out.

PR #52, opened 2019-03-25
Kokkos optimizations

stanmoore1:
## Purpose

Optimize performance for small systems running on GPUs using Kokkos. Reduce view initialization and the number of kernel launches.

## Author(s)

Stan Moore (Sandia)

## Backward Compatibility

No Issues.


stanmoore1:
Regression tests passed.

PR #78, opened 2019-06-05
Add MPIIO option for restart files

stanmoore1:
## Purpose

Add an MPIIO option for restart files, which allows multiple MPI ranks to write to one big restart file instead of many small files. Code adapted from LAMMPS.

## Author(s)

Stan Moore (Sandia)

## Backward Compatibility

No issues


stanmoore1:
Closing for now.

PR #86, opened 2019-08-29
Some performance optimizations for Kokkos package

stanmoore1:
## Purpose

Reduce memory transfer and initialization time on GPUs.

## Author(s)

Stan Moore (Sandia)

## Backward Compatibility

No issues


stanmoore1:

PR #94, opened 2019-09-24
Remove Kokkos hash in compute_surf_kokkos

stanmoore1:
## Purpose

Remove Kokkos hash in compute_surf_kokkos because it uses too many registers on GPUs and slows down performance.

## Author(s)

Stan Moore (SNL), issue reported by Gabriele Jost (NASA Ames Research Center)

## Backward Compatibility

No issues


stanmoore1:

PR #109, opened 2020-01-08
Fix thread-safety bug in Kokkos versions of compute boundary and compute surf

stanmoore1:
## Purpose

Fix thread-safety bug in `compute_boundary_kokkos` and `compute_surf_kokkos` that could lead to incorrect tallies when using GPUs or more than 1 thread.

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

No issues.


stanmoore1:
Regression tests are running.

stanmoore1:
Regression tests passed.

Issue #105, opened 2019-12-13
Read_surf calls unititialized Modify


stanmoore1:
## Summary

`read_surf.cpp:199` calls `grid_surf:903` which calls `modify:213`. The issue is that the Modify class is only initialized for a `run` command, not for `read_surf`, so `n_pergrid` is uninitialized leading to segmentation fault crash.

stanmoore1:
Fixed by #110

PR #110, opened 2020-01-08
Initialize Modify n_pergrid in constructor as a workaround for #105

stanmoore1:
## Purpose

Initialize Modify `n_pergrid` in constructor as a workaround for #105

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

No issues.


stanmoore1:
Closes #105

PR #138, opened 2020-03-20
Fix segfault issue in collide.cpp

stanmoore1:
## Purpose

`read_surf` calls `collide.cpp:1692`. The issue is that the Collide class is only initialized for a run command, not for read_surf, so `ngroups` is zero in the grow() command, leading to segmentation fault crash. This is similar to #105.

## Author(s)

Stan Moore (SNL), reported by Savio Poovathingal (University of Kentucky)

## Backward Compatibility

No issues.

### DIFF
--- a/src/collide.cpp
+++ b/src/collide.cpp
@@ -1688,7 +1688,7 @@ void Collide::adapt_grid()
 void Collide::grow_percell(int n)
 {
   if (nglocal+n < nglocalmax || !ngroups) return;
-  nglocalmax += DELTAGRID;
+  while (nglocal+n < nglocalmax) nglocalmax += DELTAGRID;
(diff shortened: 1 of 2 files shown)


src/collide.cpp
```diff
@@ -1687,7 +1687,7 @@ void Collide::adapt_grid()
 
 void Collide::grow_percell(int n)
 {
-  if (nglocal+n < nglocalmax) return;
+  if (nglocal+n < nglocalmax || !ngroups) return;
   nglocalmax += DELTAGRID;
```
sjplimp: Also, I think the 2nd line needs to be:
while (nglocal+n < nglocalmax) nglocalmax += DELTAGRID;

sjplimp:
good catch on this Stan - I was puzzling about this one ...

Issue #151, opened 2020-05-13
May 7 release read_restart crashed with more than 1 process


zhangchonglin:
## Summary

_In May 7 release, the `read_restart` command crashed during restart file reading_

## Type of Issue

_Bug Report_

## Detailed Description

Running the `May 7` release with the `circle` example (slightly modified to write and read restart file) in the `examples` folder, the code crashed when reading the restart file, with the following error message:

```
SPARTA (7 May 2020)
Reading restart file ...
  restart file = 7 May 2020, SPARTA = 7 May 2020
  orthogonal box = (0 0 -0.5) to (10 10 0.5)
ERROR on proc 0: Invalid flag in peratom section of restart file (../read_restart.cpp:199)
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
```

The code crashed with 4 processes. When using only 1 process, the code did not crash. With `Jan 24` release (305bb40), the code is running fine, successfully reading the restart file and continuing the simulation, with different number of MPI processes.

## SPARTA Version

_May 7 2020 release, with commit tag: 5e34543_

## Expected Behavior

_The code should successfully read the restart file and continue the simulation_

## Actual Behavior (Bug Report)

_Code crashed during restart file reading_

## Steps to Reproduce (Bug Report)

The files are attached. To reproduce the bug, do the followings:
- The `circle` example in the `examples` folder is slightly modified to repeat the bug. 
- `mpirun -np 4 spa_mpi -in in.circle`: this will run for 1000 steps, and create restart files: `circle.restart.500` and `circle.restart.1000`
- `mpirun -np 4 spa_mpi -in in.circle_restart`: this will reproduce the bug. It should read the restart file and continue the simulation for additional 2000 steps.

## Further Information, Files, and Links

Files to reproduce the bug are attached here:
[circle_restart.tar.gz](https://github.com/sparta/sparta/files/4622567/circle_restart.tar.gz)

stanmoore1:
The latest patch broke restart backwards compatibility in order to support writing restart files larger than 2 GB per MPI rank. However, it looks like you are writing and then reading with the same version of the code, so perhaps a bug was introduced. I'll take a look.

stanmoore1:
I can reproduce this issue, should have a fix soon.

zhangchonglin:
Thanks for the quick confirmation!

zhangchonglin:
Thanks for the quick fix. I can confirm the crash issue is fixed now.

[...]

PR #152, opened 2020-05-13
Fix issue in restart file read

stanmoore1:
## Purpose

Fix issue in restart file read that causes crashes in certain cases. This was due to #134.

## Author(s)

Stan Moore (SNL), reported by @zhangchonglin

## Backward Compatibility

No issues.


stanmoore1:
Fixes #151.

stanmoore1:
I tested all the branches in the code, hopefully I found all the issues.

PR #122, opened 2020-02-04
added optimized particle moves (work in progress) and removed execute…

akstagg:
… permissions on several source and header files.

## Purpose
This pull request features the addition of optimized particle moves through addition of "global   optParticleMoves" to the input.  Default behavior is the unoptimized move.

## Author(s)
Alan Stagg

## Backward Compatibility
Backward compatability is unaffected.

## Implementation Notes
in progress

## Post Submission Checklist

_Please check the fields below as they are completed_
- [ ] The feature or features in this pull request is complete
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] One or more example input decks are included
- [ ] The source code follows the SPARTA formatting guidelines

## Further Information, Files, and Links

_Put any additional information here, attach relevant text or image files, and URLs to external sites (e.g. DOIs or webpages)_

PR #156, opened 2020-05-14
Add optimized particle moves

stanmoore1:
## Purpose

This pull request features the addition of optimized particle moves through addition of `global optmove` to the input. Default behavior is the unoptimized move.

## Author(s)

Alan Stagg (SNL), Stan Moore (SNL)

## Backward Compatibility

No issues.


stanmoore1:
This PR replaces #122.

src/update.cpp
```diff
@@ -279,6 +267,36 @@ void Update::run(int nsteps)
 
     // move particles
 
+    if (grid->uniform && enable_optmove)
```
sjplimp, line 270: I think this should be set once before the run (e.g. in init), not every step, whcih is when the error or warning is issued.  As it is, this will be set even if surfs exist - won't tjhat be an error in comm.cpp ?

stanmoore1: @sjplimp the thought was fix adapt could change the grid from uniform to non-uniform, which could happen mid-run, so we need to check for that every timestep.

stanmoore1: I suppose we could just error out if fix adapt is active and a user enables optimized moves though.

sjplimp: That's what I suggest.  I don't think we should mix opt move with grid adaptation.  If I understand it, opt move is for systems where particles uniformly fill a box.  No surfs, limited density variations.  If a user wants to create a coarse gird, then adapt it to a system with a lot of particles, they can do that in the first run, then enable opt move and do a second (equilibrium) run.

src/update.cpp
```diff
@@ -333,13 +351,26 @@ template < int DIM, int SURF > void Update::move()
...
+  double Lx,Ly,Lz,dx,dy,dz;
+  double *boxlo, *boxhi;
   Grid::ParentCell *pcell;
   Surf::Tri *tri;
   Surf::Line *line;
   Particle::OnePart iorig;
   Particle::OnePart *particles;
   Particle::OnePart *ipart,*jpart;
 
+  if (enable_optmove) {
```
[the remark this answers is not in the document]
stanmoore1: Yes good catch.

stanmoore1:
I addressed @sjplimp's comments.

PR #161, opened 2020-06-03
Fix issues in low-memory restart file read/write

stanmoore1:
## Purpose

Fix issues in low-memory restart file read/write when a file has zero particles and when using custom option for particles.

## Author(s)

Stan Moore (SNL), reported by Michael Gallis (SNL)

## Backward Compatibility

No issues.

### DIFF
--- a/src/particle.cpp
+++ b/src/particle.cpp
@@ -1459,9 +1459,10 @@ int Particle::pack_restart(char *buf)
    use multiple passes to reduce memory use
    use OnePartRestart data struct for permanent info and to encode cell ID
    include per-particle custom attributes if defined
+   NOTE: does not ROUNDUP(ptr) at the end
 ------------------------------------------------------------------------- */
 
-int Particle::pack_restart(char *buf, int step, int pass)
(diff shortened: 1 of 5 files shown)


src/particle.cpp
```diff
@@ -1541,9 +1537,10 @@ int Particle::unpack_restart(char *buf)
    use multiple passes to reduce memory use
    allocate data structure here, will be deallocated by ReadRestart
    include per-particle custom attributes if defined
+   NOTE: does not ROUNDUP(ptr) at the end
```
sjplimp, line 1540: add: ... end, this is done by caller

src/particle.cpp
```diff
@@ -1459,9 +1459,10 @@ int Particle::pack_restart(char *buf)
    use multiple passes to reduce memory use
    use OnePartRestart data struct for permanent info and to encode cell ID
    include per-particle custom attributes if defined
+   NOTE: does not ROUNDUP(ptr) at the end
```
sjplimp, line 1462: ditto to my other comment

PR #189, opened 2020-10-09
tools/paraview/grid2paraview: Fix bugs

timteichmann:
## Purpose

This pull request fixes two issues in the grid2paraview python tool.
commit 1:
Fix bug where using the command `pvpython grid2paraview.py my_grid sparta2pv` would result in unreadable `.pvd` files if no time grid data file(s) were provided

commit 2:
Fix python `ValueError "Mixing iteration and read methods would lose data."` that occurs when `readline()` is used in a for loop over the lines.

## Author(s)

Tim Teichmann (Institute for Technical Physics, Karlsruhe Institute of Technology)

## Backward Compatibility

should not have any issues

## Implementation Notes

Script does not run at all without 2nd commit (see above, `ValueError`).  Applying the 2nd commit makes the script execute without error with `pvpython` from ParaView 5.4.
However, even then if no time grid data files are provided via the `-r ...` option the script will produce an unreadable `.pvd` files (this happens because the previous `else: return` gets called and the footer of the `.pvd` file is never written). The 1st commit then makes the script produce sane paraview `.pvd` files even if no time grid data is provided.

## Post Submission Checklist

### DIFF
--- a/tools/paraview/grid2paraview.py
+++ b/tools/paraview/grid2paraview.py
@@ -928,18 +928,16 @@ def write_pvd_file(time_steps_dict, file_name, num_chunks):
       fh.write('    <DataSet timestep="' + str(time) + '" group="" part="0"   \n')
       filepath = os.path.join(file_name, file_name + '_'  + str(time) + '.pvtu')
       fh.write('             file="' + filepath + '"/>\n')
+      array_names = []
       file_list = time_steps_dict[time]
       if file_list:
         try:
           afh = open(file_list[0], "r")
+          read_array_names(afh, array_names)
+          afh.close()
         except IOError:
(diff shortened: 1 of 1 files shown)


stanmoore1:
@tjotaha was this fixed by #190, otherwise can you take a look?

tjotaha:
@stanmoore1 It looks like only the second commit of #189 was fixed by #190 for the "Mixing iteration .. " error. The first commit in #189 about the unreadable *.pvd files for no grid time data was not addressed by #190

timteichmann:
Hi,
#190 did the same thing the second commit did here (just the other way round basically..)
I removed that commit. The second commit stays open as commented by @tjotaha

stanmoore1 (approved):

Issue #150, opened 2020-05-09
Particle sent to self error


aborner1:
## Summary

Frequent encounter of the "Particle sent to self" error in axisymmetric calculations.

## Type of Issue

Bug Report

## SPARTA Version (Bug Report)

Current master but not a new bug.

## Expected Behavior (Bug Report)

For axisymmetric simulations, I as well as other collaborators have frequently encountered the "ERROR on proc xxx : Particle xxxx on proc xxx being sent to self on step xxxxxx (../update.cpp:1167)". This error should never happen.

## Steps to Reproduce (Bug Report)

It happens more frequently for larger simulations on a large number of processors, but working on finding a simpler test case that reproduces the bug so it can be shared. It is, however, random number dependent.

## Further Information, Files, and Links

I propose to turn the error into a warning with a printout (until the bug can be understood and fix), as it will only affect 1 or 2 particles out of billions, and delete those particles rather than end the simulation.

At the location indicated in the bug report in update.cpp, replace the following lines:
          if (cells[icell].proc == me) {
            char str[128];
            sprintf(str,
                    "Particle %d on proc %d being sent to self "
                    "on step " BIGINT_FORMAT,
                    i,me,update->ntimestep);
            error->one(FLERR,str);
          }
 
By:
          if (cells[icell].proc == me) {
//            char str[128];
            printf("Particle %d on proc %d being sent to self on step %d: DISCARDED\n",
                   i,me,ntimestep);
            particles[i].flag = PDISCARD;
//            error->one(FLERR,str);
          }

stanmoore1:
In a different code, we also had issues with axisymmetric. It seems to be more prone to round error issues. @sjplimp and @mgallis, what do you think?

stanmoore1:
I talked to @sjplimp about this and he would like to understand *why* this is happening. @aborner1 do you have a reproducer input script?

aborner1:
I only have large cases that reproduce it, I would need to see if I can generate something small that ideally can be run on a small number of procs, otherwise it's going to be a nightmare to debug.

stanmoore1:
@aborner1 I have assigned this issue to you. Once we have a reproducer I will reassign it to @sjplimp. Thanks.

stanmoore1:
This issue is possibly fixed by #194, @aborner1 can you confirm?

aborner1:
[...]

PR #194, opened 2020-11-19
Fix very rare bug in axisymmetric

stanmoore1:
## Purpose

In axisymmetric (only), if a particle lands *exactly* (within roundoff error) on a cell boundary, and has a negative y-velocity, then it can be assigned the wrong cell, which can lead to crashes. 

## Author(s)

Steve Plimpton, Stan Moore (SNL)

## Backward Compatibility

No issues.


stanmoore1:
@aborner1 there is a high chance this PR will fix #150.

PR #206, opened 2020-11-26
Fix mflux

timteichmann:
## Purpose

Currently the computation of the mass flux ("mflux") in compute_{surf,isurf,boundary} is flawed in two ways:

1. in case surface reactions (exchange or dissociation) are involved the wrong fields of the array are updated as the index is incremented erroneously. This pull request remedies this
2. for {surf,isurf}: Contrary to the descriptions in the code and on the manual the commands currently do not calculate a flux but only the sum of the masses of the impinging DSMC particles. This pull request adds the fluxscale to all computations (like for example in case of the kinetic energy).

## Author(s)

Tim Teichmann (Institute for Technical Physics, Karlsruhe Institute of Technology)

## Backward Compatibility

Yes, since results of the involved computes change

## Implementation Notes

- Commit 1: Do not increment the index if surface reactions are considered in the mass flux
- Commit 2: Scale the mass with the fluxscale (FNUM/A/dt) to obtain the mass flux

## Post Submission Checklist

I think the changes are relatively straightforward. However, please let me know if you would like me to upload a small case to demonstrate the effect of these changes.

NOTE: I can currently not test the KOKKOS version. But I do not expect any issues with them.

### DIFF
--- a/src/KOKKOS/compute_boundary_kokkos.h
+++ b/src/KOKKOS/compute_boundary_kokkos.h
@@ -168,9 +168,10 @@ void boundary_tally_kk(int iface, int istyle, int reaction,
         a_myarray(iface,k++) += weight;
         break;
       case MFLUX:
-        a_myarray(iface,k++) += origmass;
-        if (ip) a_myarray(iface,k++) -= imass;
-        if (jp) a_myarray(iface,k++) -= jmass;
+        a_myarray(iface,k) += origmass;
+        if (ip) a_myarray(iface,k) -= imass;
+        if (jp) a_myarray(iface,k) -= jmass;
+        k++;
         break;
       case PRESS:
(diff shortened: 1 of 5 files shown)


aborner1:
I concur with these changes. I have noticed both issues in the past.

stanmoore1:
@timteichmann I agree with these changes. However, the logic is a little confusing. Can you put the `k++` at the very end so that it uses `k` for all 3 indices instead of `k-1` for the last two?

timteichmann:
Agreed, that makes it much clearer. Should be fine now?

stanmoore1 (approved):

stanmoore1:
@sjplimp can you take a look at this PR? It looks correct to me.

sjplimp (approved):

sjplimp:
good catches @timteichmann - thanks!

PR #210, opened 2020-12-16
fix_emit: use MY_PIS instead of sqrt(MY_PI)

timteichmann:
## Purpose

fix_emit used `sqrt(MY_PI)` instead of the already existent constant `MY_PIS`. Since it is potentially faster to use the constant directly this pull request changes it. This is the only place in the code where this is the case (checked by `git grep "sqrt(MY_PI)"`).

## Author(s)

Tim Teichmann (KIT, ITEP)

## Backward Compatibility

no issues

## Implementation Notes

Since the constant exists make use of it to potentially speed up the simulation. I don't know how sensitive your regression tests are. This might trigger some due to small numerical differences.

## Post Submission Checklist


stanmoore1 (approved):

stanmoore1:
@timteichmann nice catch

PR #211, opened 2020-12-17
grid: do not include axial cell size in radial weight computation

timteichmann:
## Purpose

Currently the computation of radial weights is somewhat unexpected, at least in my opinion. Based on the explanation in the doc (https://sparta.sandia.gov/doc/global.html)

>  The radius setting is only allowed for axisymmetric systems. The weight in this case is the distance the cell's midpoint is from the y=0 axis of symmetry.

the user expects that `global weight radius` will generate cell weights solely dependent on the radial position of the cell (i.e. `weight=0.5(yhi+ylo)`). What is actually done is `weight=0.5(yhi+ylo) * (xhi-xlo)`, i.e. the axial extension of the cell is also taken into account. In my opinion this does not make a lot of sense. In a good axisymmetric mesh the cell size itself should be adjusted to the local density, therefore ensuring an approximately equal number of simulator particles in each cell (e.g. in the axisymmetric blunt cylinder calculations).

If the current behaviour was implemented intentionally, I would suggest to add a short comment on the actual behavior in the docs.

If this is merged the example for axisymmetry should probably also be adjusted. While the grid in the example does not feature different cell sizes in x direction the FNUM value should be updated from `1e17` to `1e17*0.025` to keep the effective FNUM value the same. If you want to merge this I can add it here.

## Author(s)

Tim Teichmann (KIT, ITEP)

## Backward Compatibility

**This is not backward compatible.** For a typical grid with cell dimensions in the µm range this will change the effective FNUM value by a factor of 1e6. Input scripts will therefore have to be adjusted and previously generated restart files will not continue as expected!

## Implementation Notes

Change the computation of the weight from
`weight=0.5(yhi+ylo) * (xhi-xlo)`
to
`weight=0.5(yhi+ylo)`

## Post Submission Checklist


stanmoore1:
@timteichmann I see your point, but I've assigned @sjplimp as he should know if there is a reason the weight was defined with `(xhi-xlo)`.

sjplimp (approved):

sjplimp:
@timteichmann I agree with this change.  The weight values are used as a ratio between the final cell a particle moves to versus its initial cell, to clone or delete particles.  Thus I don't think this change will make much difference for particle moves, except for adapted grids where particles move in x from one size grid cell to another.  In which case the change is correct.

However the weight is also used in fix emit/face and emit/surf to set the number of particles to insert.  Which is why you noted the big change in effective fnum.   I'm thinking in this case that when the target insertion number is scaled by 1/weight this is now the wrong dimensionality (radial distance instead of area).  So that for the same fnum you would end up running a problem with dramatically different particle count.  I don't think that is right, or that simply changing fnum is correct.  I'm thinking maybe the xwidth factor needs to be added back in for that computation?  Any thoughts on this?

sjplimp:
@stamoor please hold off on the merge until the fix emit issue in my previous comment is figured out

stanmoore1:
@sjplimp will do. Please assign this PR to me when it is ready for merge.

timteichmann:
@sjplimp Do you mean e.g. in `fix_emit_surf.cpp`:

`ntargetsp *= nrho*area*dt / fnum;`
`ntargetsp /= cinfo[icell].weight;`
?

So if we transfer to a simple 2D case `cinfo[icell].weight = 1` for all cells, no matter what dimensions they have, so a cell with dx,dy=4 would have the same weight as a cell with dx,dy=2. For this reason I would argue that the same holds true in the axisymmetric case? The different weight should only come from the radial position so as to reduce the number of particles inserted at radii further away from the axis? I might think about this the wrong way though but maybe this outlines my thinking better.

sjplimp:
yes, that's the code.  Similar in fix emit/face.  Take emit/face as the example.  My thinking is that if the user chooses an fnum, then for both regular 2d and 2d-axi (or 3d for that matter), there is a statistically correct number of particles that should be in the simulation box, e.g. assuming steady-state flow and no surface objects.  For 2d-axi the number density of particles is non-uniform in the radial direction, but there is still a "correct" distribution and correct total count that matches fnum.  I think fix emit/face should inject particles that match that correct distribution for 2d-axi.

I think you are saying that dividing by the cell's y-position will produce that?  Which would mean that previously the code was not doing that.  I.e. fix emit/face was injecting the (sometimes wildly) wrong number of particles for 2d-axi?

timteichmann:
I think the latter for really bad meshes (probably not for any real production mesh). Note that this does not mean that the solution will be wrong. As the weight can basically be chosen at free will, it will still produce the correct number of collisions etc.. So if sufficient particles are in each cell the solution should stay the same.

Imagine an axisymmetric case of flow through a pipe. Lets say the flow is from left to right (x-direction) with inflow boundary on the left. For simplicity we have only two cells, cell 1 and cell 2, at the inlet, both with equal dy, dx, so that dy = 1/2 * r (i.e. the cells are stacked in radial direction). Now everything is fine: We would inject more particles in the cell 2 without radial weighting but if we switch on radial weighting approx. equal numbers of particles are injected in cells 1 and 2.

Now imagine that we split cell 1, lets call the split cells 1.1 and 1.2, in two according to dx_new = 1/2 dx and dy_new = dy. I.e. the face through which we inject stays exactly the same and now belongs to cell 1.1. Cell 1.2 does not lie on the boundary therefore nothing is injected here (correct). What I would expect is that approx. the same number of particles are injected in the new cell 1.1 and cell 2 (which stayed at dimensions dx, dy). However, the current version of Sparta would emit only **half** the number of particles in the new (smaller) cell 1.1 because the weight is also scaled with dx.

If we now switch back to 2D (so imagine the axis as the symmetry plane in a 2d channel) we would inject identical number particles in cell 1 and cell 2 (before split of cell 1) and cell 1.1 and cell 2 (after split of cell 1).

In my opinion the latter case is correct and should also apply to axisymmetric cases.

(Note that I tried to include a PNG of the lengthy explanation above but Github somehow fails to upload all the time..)

sjplimp:
ok - thanks for the careful explanation - makes sense.   @stanmoore1 I think is now ready to merge. Re: the example script, you can reset fnum = 2.5e15 to keep the particle count roughly the same.  But don't worry about an exact match to previous result.  You can just re-bless the output for the regression testing.

timteichmann:
Shall I push a commit for the example?

stanmoore1:
Thanks @timteichmann I'll update the ref log.

timteichmann:
Great thanks :)

stanmoore1:
@timteichmann I don't have permission to push into your branch, so I submitted a PR.

timteichmann:
@stanmoore1 Oh woops, sorry. It is merged.

stanmoore1 (approved):

PR #214, opened 2020-12-24
Flip sign of internal FFTs

sjplimp:
## Purpose

Flip the sign internally for performing forward vs backward FFTs.  This is to be compatible with other FFT libraries.  It should not affect results for the compute fft/grid command when the conjugate flag is set (e.g. for energy spectra).  If the result of one or more FFTs is output directly, then the sign of the imaginary term will be flipped.

## Author(s)

Steve

## Backward Compatibility

See above.

## Implementation Notes

## Post Submission Checklist

### DIFF
--- a/src/FFT/fft2d.cpp
+++ b/src/FFT/fft2d.cpp
@@ -91,23 +91,23 @@ void fft_2d(FFT_DATA *in, FFT_DATA *out, int flag, struct fft_plan_2d *plan)
   length = plan->length1;
 
 #if defined(FFT_MKL)
-  if (flag == -1)
+  if (flag == 1)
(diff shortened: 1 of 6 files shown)


src/FFT/fft2d.cpp
```diff
@@ -440,14 +423,6 @@ struct fft_plan_2d *fft_2d_create_plan(
                             NULL,&nslow,1,plan->length2,
                             FFTW_BACKWARD,FFTW_ESTIMATE);
 
-  if (scaled == 0)
```
stanmoore1, line 443: @sjplimp why is this code removed?

stanmoore1 (approved):
I had one question, looks good.

PR #221, opened 2021-01-15
Flag particles for deletion with rare axisymmetric move round-off issues

sjplimp:
## Purpose

Add logic to remove particles which end up epsilon outside a curved cell surface in axisymmetric cells at the end of the move, when the move logic thinks they are inside.  In final run stats, print the number of such removals.

## Author(s)

Steve

## Backward Compatibility

_Please state whether any changes in the pull request break backward compatibility for inputs, and - if yes - explain what has been changed and why_

## Implementation Notes

## Post Submission Checklist

### DIFF
--- a/src/KOKKOS/update_kokkos.cpp
+++ b/src/KOKKOS/update_kokkos.cpp
@@ -83,8 +83,8 @@ UpdateKokkos::UpdateKokkos(SPARTA *sparta) : Update(sparta),
 
   // use 1D view for scalars to reduce GPU memory operations
 
-  d_scalars = t_int_11("collide:scalars");
-  h_scalars = t_host_int_11("collide:scalars_mirror");
+  d_scalars = t_int_12("collide:scalars");
+  h_scalars = t_host_int_12("collide:scalars_mirror");
 
   d_ncomm_one     = Kokkos::subview(d_scalars,0);
   d_nexit_one     = Kokkos::subview(d_scalars,1);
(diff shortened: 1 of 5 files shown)


stanmoore1:
@sjplimp do we want to have a second option that searches for the correct cell and puts the particle back in, or do you think just bad deleting the particles is sufficient?

sjplimp:
@stanmoore1 I think that becomes too complicated.  B/c of parallelism (which proc owns that cell).  And b/c a surf could lie along the boundary of the 2 cells.  Assuming it is very rare, I think just deleting the offending particle is ok.

mgallis:
For open systems deleting the stray particle should be sufficient.
For closed systems, placing it back in allows for mass conservation to be satisfied, although artificial average velocities pointing to the wall may appear.

Michael

On Jan 15, 2021, at 11:43 AM, Stan Moore <notifications@github.com> wrote:

﻿

@sjplimp<https://github.com/sjplimp> do we want to have a second option that searches for the correct cell and puts the particle back in, or do you think just bad deleting the particles is sufficient?

stanmoore1:
@sjplimp this does indeed allow the reproducer to run farther, and caught the bad particle:

```
Particles stuck   = 0
Axisymm bad moves = 1
```

stanmoore1:
I'm porting to Kokkos now.

stanmoore1 (approved):

stanmoore1:
I ran the reproducer for 1.38T particle moves (over 2 hours runtime on 256 cores) and there was only a single bad axisymmetric move, so this seems to be a really rare issue.

PR #234, opened 2021-02-22
Fix bug in surf grow() call

stanmoore1:
## Purpose

Fix bug in a surf `grow()` call where the old value was larger than the new value, leading to a negative input into memset, causing a crash in the `/heavy/corner_test/in.corner.distributed` regression test.

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

Yes.


sjplimp (approved):

stanmoore1:
@sjplimp would it be better to move this check inside `Surf::grow()` in `surf.cpp` instead of in `grid_surf.cpp`?

sjplimp:
yes, that would be fine, and safer.

PR #236, opened 2021-02-23
Preserve previous axisymmetric weighting scheme as an option

stanmoore1:
## Purpose

Preserve previous axisymmetric weighting scheme as an option. This was removed in PR #211 but according to Michael Gallis it may have advantages in some cases.

## Author(s)

Stan Moore, Steve Plimpton, Michael Gallis (SNL)

## Backward Compatibility

This reverts back to the old axisymmetric weighting scheme in the `axi/in.axi` example, but the new weighting scheme as in #221 is also an option.


stanmoore1:
@sjplimp The `radius/unweight` command name could be changed if you want, I just threw something out there as a place holder.

sjplimp:
@stanmoore1 Just changed the syntax and doc page a bit.  Hopefully didn't break anything.

PR #242, opened 2021-03-07
Small typos fix in Sparta Python documentation

zhangchonglin:
## Purpose

_While looking at the Sparta Python interface documentation, I found these small typos_

## Author(s)

_Chonglin Zhang, Rensselaer Polytechnic Institute_

## Backward Compatibility

_No, since only documentation typos fix_

## Implementation Notes

_Modified typos and broken links in the documentation._

### DIFF
--- a/doc/Section_python.html
+++ b/doc/Section_python.html
@@ -103,7 +103,7 @@ will load by default.  Note that if you are building multiple machine
 versions of the shared library, the soft link is always set to the
 most recently built version.
 </P>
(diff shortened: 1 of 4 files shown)


stanmoore1:
@zhangchonglin good catch.

@sjplimp It looks like `make -f Makefile.shlib mpi` is broken, I get `cp: cannot stat ‘fix_inflow.cpp’: No such file or directory`. `Makefile.shlib` seems to be out of date, can this file be removed, or should we update it? Also not sure that `Makefile.shlib` would work with KOKKOS.

stanmoore1 (approved):

PR #249, opened 2021-05-13
Update Kokkos library in SPARTA to v3.4.0

stanmoore1:
## Purpose

Update Kokkos library in SPARTA to v3.4.0

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

Requires CMake v3.16 or newer.


stanmoore1:
Kokkos regression tests passed and I checked the performance of the `collide` benchmark on on V100.

Issue #187, opened 2020-09-04
Inelastic collision problems in SPARTA code


realprivate:
## Summary

The issue discusses the inelastic collision bugs in SPARTA code, and it also puts forward a proposal for the code improvement.

## Type of Issue

Report bugs, and also give an enhancement suggestion

## Detailed Description (Enhancement Suggestion)

Up to now, there are still some bugs in SPARTA code (most probable in collide_vss.cpp) about the inelastic collision process. For instance, when simulating molecular gases in equilibrium (see attachments), setting the relax value to "variable" will cause the rotational temperature to be obviously wrong, and setting the vibrate value to "discrete" will cause the vibrational temperature to be obviously wrong. Only when setting the relax value to "constant", and vibrate value to "smooth" will lead to the correct result.

Besides, the SPARTA now still uses the serial implementation from Bird to handle non-reactive inelastic collisions. In this case, if both particles are selected for an inelastic collision, there is some degree of coupling between their relaxation process. Furthermore, when both rotational and vibrational relaxation processes are considered, sequential testing for rotational followed by vibrational inelastic collisions will inherently couple the rotational and vibrational relaxation processes. In order to reproduce the specified internal energy relaxation rates for a given energy mode and species interaction, it would be better if SPARTA used the particle selection method prohibit double relaxation (from Haas et al, 1994), to replace the current method.

## SPARTA Version (Bug Report)

It is the latest SPARTA version, sparta-6Jul20.

## Expected Behavior (Bug Report)

When simulating molecular gases in equilibrium, the final translational temperature Ttran, rotational temperature Trot and vibrational temperature Tvib should be approximately equal, regardless of how the relaxation parameters are set in the input file. 

## Actual Behavior (Bug Report)

When setting relax value to "variable", the Trot will be obviously smaller than Ttran and Tvib. And when setting vibrate value to "discrete", the Tvib will be obviously smaller than Ttran and Trot. Only when setting the relax value to "constant", and vibrate value to "smooth" will lead to the right result. 

## Steps to Reproduce (Bug Report)

The attachments below are the input files of the test case, which simulates the O2 gas in equilibrium. Changing the "collide" and "collide_modify" command will reproduce the problems above. The filename should be changed (air.vss, in.relax, air.species) before testing.

## Further Information, Files, and Links

[air_vss.txt](https://github.com/sparta/sparta/files/5172533/air_vss.txt)
[in_relax.txt](https://github.com/sparta/sparta/files/5172535/in_relax.txt)
[...]

PR #268, opened 2021-06-24
Rotrel equilibrium fix

### LINKED ISSUE #2: Variable Rotational Relaxation

zseckert:
## Purpose

These changes address the variable rotational portion of #187.

It was brought to our attention by Maninder Grover and Dr. Paolo Valentini that including the rotational energy of the relaxing particle in the call to rotrel resolved the issue with temperature varying rotational relaxation not maintaining equilibrium at steady state.

Eq (3.32) in Bird 2013 discusses the same inclusion, with the added detail of including the rotational DoF in the calculation for T_coll. There are discussions of shortcomings of this approach related to partially excited modes, but we feel the steady state equilibrium issue is much more significant.

Further analysis is being performed to understand exactly why the non-equilibrium is not maintained with the original implementation and to hopefully provide insights to fixes of the issues with the current implementation discussed by Bird.

## Backward Compatibility

This will change the results of simulations including temperature varying rotational relaxation

## Implementation Notes

This does deviate from the original Parker (1959) derivation of the temperature dependence of Zrot, but our tests examining both the relaxation rate and the rotational energy distribution function show no significant deviations from what is expected.

## Post Submission Checklist

There is currently not an example that uses variable relaxation, but I can add one based on collide if desired.

Fixes #2.


stanmoore1:
@zseckert thanks, I will update the Kokkos version.

stanmoore1 (approved):

stanmoore1:
>There is currently not an example that uses variable relaxation, but I can add one based on collide if desired.

@zseckert This would be very good, if possible. All the regression tests passed, which confirms this case isn't covered in our test suite.

zseckert:
> @zseckert This would be very good, if possible. All the regression tests passed, which confirms this case isn't covered in our test suite.

Okay, I will do so. Is there a way to run the regression tests locally? It looks from the results that it is CTest based, so is using the CMake build sufficient?

stanmoore1:
Yes the tests use CTest, instructions are here: https://sparta.github.io/doc/Section_start.html#start_5.

zseckert:
I have added two tests. The first is for constant Zr that equilibrates to the expected 6040 K for both T_tr and T_r.  The second uses variable Zr, and without the fix, T_r equilibrates about 200 K below T_tr. With the fix, they are within a couple K of the expected.

Both examples are on 3x3x3 grids to avoid the single cell/serial bug of equilibrium not being reached, but they run fast enough on 1 cell/1 core (about 90 s) that once that fix is complete, the constant Zr problem could also act as regression for that (It currently has about 200 K non-equilibrium when run in 1 cell).

zhangchonglin:
@zseckert: I think this is more of an issue with regard to equipartition of energy between different energy modes, when using the Borgnakke–Larsen model to sample the post-collision energy. Usually the relaxation rate was not affected much. Simply speaking, within the Borgnakke-Larsen model framework and using energy dependent collision probability for inelastic collision (in this case the Parker model), we want to sample a `"biased"' energy distribution function (the energy distribution from all molecules selected for inelastic collision), to ensure equipartition of energy modes.

This issue was first discussed and resolved by Abe in his RGD paper: `“Direct Simulation Monte Carlo Method for Internal- Translational Energy Exchange in Nonequilibrium Flow,” Rarefied Gas Dynamics: Theory and Simulations; Proceedings of the 18th International Symposium on Rarefied Gas Dynamics, Univ. of British Columbia, Vancouver, BC, Canada, 1994, pp. 103–113`. Since this is some implementation details that is usually overlooked, I had some discussions of this in our AIAA journal paper: Nonequilibrium-Direction-Dependent Rotational Energy Model for Use in Continuum and Stochastic Molecular Simulation. Kosyanchuk and Yakunchikov also have some excellent discussions of this in their recent Physics of Fluids paper, from a different perspective: `A detailed multiscale study of rotational–translational relaxation process of diatomic molecules`.

stanmoore1:
Thanks @zseckert. I'm going to run the example scripts in #187 and #2 to see if the equilibrium issues are now resolved.

mgallis:
You may want to use the new RNG Stan implemented.

Michael

On Jun 25, 2021, at 7:58 AM, Stan Moore ***@***.***> wrote:

﻿

Thanks @zseckert<https://github.com/zseckert>. I'm going to run the example scripts in #187<https://github.com/sparta/sparta/issues/187> and #2<https://github.com/sparta/sparta/issues/2> to see if the equilibrium issues are now resolved.

stanmoore1:
This does indeed fix #2.

stanmoore1:
I also confirm this fixes the issue shown by the input script in #187.

stanmoore1:
To clarify, this fixes the variable rotational portion of #187 as @zseckert mentioned in the description. Using `vibrate discrete` in the input in #187 still gives vibrational nonequilibrium.

sjplimp (approved):

PR #271, opened 2021-06-30
Add warning related to computing temperature and fix vibmode

stanmoore1:
## Purpose

Add warning if using compute grid tvib with fix vibmode, which can give incorrect temperature.

## Author(s)

Stan Moore (SNL), reported by Arnaud Borner (NASA)

## Backward Compatibility

Yes


sjplimp (changes requested):
I think this check needs to be in the init() of ComputeGrid.  B/c fix vibmode could be defined after compute grid is defined.

stanmoore1:
> I think this check needs to be in the init() of ComputeGrid. B/c fix vibmode could be defined after compute grid is defined.

Good point, done.

PR #272, opened 2021-07-08
Add nsplit to dump_grid

timteichmann:
## Purpose

Add `nsplit` to `dump_grid` command. This is useful to make dump of split cells (which have identical cell ids) unique and is easier to compare against than the volume which was the best previously available option.

## Author(s)

Tim Teichmann, KIT

## Backward Compatibility

100% compatible

## Post Submission Checklist

### DIFF
--- a/doc/dump.html
+++ b/doc/dump.html
@@ -59,6 +59,7 @@
 </PRE>
 <PRE>      id = integer form of grid cell ID
       idstr = string form of grid cell ID
+      nsplit = integer, either 1 (unsplit cell) or negative sub cell ID
       proc = processor that owns grid cell
       xlo,ylo,zlo = coords of lower left corner of grid cell
       xhi,yhi,zhi = coords of lower left corner of grid cell
--- a/doc/dump.txt
+++ b/doc/dump.txt
@@ -48,6 +48,7 @@ args = list of arguments for a particular style :l
 
       id = integer form of grid cell ID
       idstr = string form of grid cell ID
+      nsplit = integer, either 1 (unsplit cell) or negative sub cell ID
       proc = processor that owns grid cell
       xlo,ylo,zlo = coords of lower left corner of grid cell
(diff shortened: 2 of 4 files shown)


stanmoore1 (approved):

stanmoore1:
Looks good to me.

sjplimp:
Conceptually this looks fine, but I think the output will be confusing to the user.  Namely ouptutring 1 for unsplit and 0,-1,-2,...
for the split cell index.  Maybe we should output 0 for an unsplit cell and 1,2,3,... for an index to the split cell.
In other words don't output the raw nsplit value as it is stored, but have dump_grid convert it to something more intelligible to someone who reads or processes the dump file.

timteichmann:
I agree that the current output is not very legible. What do you think about a rename of the dumped property to `splitid` and values of
`splitid = -1`: unsplit
`splitid = 0,...,N`: split cell index
? This would prevent a shift of the index by 1. Also the -1 can be interpreted as "NaN".

sjplimp:
I still think your suggestion is too much like a programmer would think, not a user or human reader of the file.
I prefer nsplit or split as the name (IDs are generally unique).  0 for an unsplit cell.  1,2,3,... for the index of a split cell.
These indices are the same as how columns are specified in a per-grid array by the dump grid or variable command syntax.
Does this choice somehow make it harder for post-processing the dump grid output file?

timteichmann:
> I still think your suggestion is too much like a programmer would think

True ;-). Okay I will implement it as
`split`
`0`: unsplit
`1..N`: split, where `split index = -(x-1)`

sjplimp:
Thanks! Should be ready to merge when that is finished

timteichmann:
The changes are already pushed. Let me know if anything else needs adjustment. Otherwise should be good to go.

stanmoore1 (approved):

PR #277, opened 2021-08-06
Implement per element variable surface temperature

aborner1:
## Purpose

This PR implements a variable surface temperature model based on the re-radiative equilibrium law. The surface temperature is a now a per surface element variable.

## Author(s)

Arnaud Borner (NASA Ames)

## Backward Compatibility

Should be backwards compatible.

## Implementation Notes

Implement a new fix, "fix surf/temp" that recomputes the surface temperature for each line or try, from a "fix surf", and uses the total heat flux per surface element ("etot").
This feature is currently only compatible with explicit, non-distributed surfaces. Support for distributed and implicit surfaces will be added in the future.

## Post Submission Checklist

- [X] Suitable new documentation files and/or updates to the existing docs are included (finalizing doc)


aborner1:
@stanmoore1 @sjplimp I think this PR is ready for review, I have completed the doc, example and tested it.

aborner1:
No clue why the jagged.2d.distributed test failed on 4 threads as I was able to run it locally and the log file was identical to the gold standard. From my point of view, all the tests passed.

aborner1:
I this one has an issue. I am still working on it.

aborner1:
@stanmoore1 This is ready to be reviewed and merged, I just tested it and compared to another DSMC code. I don't know why the 2 tests fail as they both pass on my Mac (in the example folder). They return identical results to the gold standard log. One test times out which is even odder.

aborner1:
@stanmoore1 @sjplimp Talked to Michael and he green-lighted this. Can we proceed with review and merge?

stanmoore1:
@aborner1 I also need to port these changes to Kokkos.

stanmoore1:
@aborner1 when I try to run the `in.sphere.adjust_temp.distributed` example I get:
```
ERROR: Cannot currently use fix surf temp with implicit or distributed surfs (../fix_surf_temp.cpp:47)
```

stanmoore1:
Should I remove that example, or is there a fix for this needed?

aborner1:
> Should I remove that example, or is there a fix for this needed?

I placed the example here to show that distributed surfaces were not yet supported, but I agree it may be confusing. Please remove if you think that is the best approach.

stanmoore1:
@aborner1 it seems like this needs to be extended to other surf collide methods (e.g. cll, impulsive, etc)?

aborner1:
> @aborner1 it seems like this needs to be extended to other surf collide methods (e.g. cll, impulsive, etc)?

From a computer science point of view, yes, but I am not sure it makes physical sense. If the wall is adiabatic and Stefan-Boltzmann law applies, then the surface should be diffuse. Now, diffuse is a sub-case of CLL, so maybe that one would make sense, but I don't know about all the other types.

sjplimp:
@aborner1 I looked this over - I can see it is a useful idea.  I think it needs to be implemented a bit differently for current SPARTA and use a per-surf custom vector.  I can make the changes if that's alright.

aborner1:
> @aborner1 I looked this over - I can see it is a useful idea. I think it needs to be implemented a bit differently for current SPARTA and use a per-surf custom vector. I can make the changes if that's alright.

Of course, that is alright with me.

sjplimp:
@aborner - this is ready for you to do some more testing on - I will email you a few final Qs

stanmoore1:
Running full regression tests now

stanmoore1:
Some of the Kokkos tests are hanging, will debug

stanmoore1:
Still seeing a hang on the GPU, debugging...

stanmoore1 (approved):

PR #278, opened 2021-08-16
Surfchem heatflux

aborner1:
## Purpose

New feature: add an option to "compute surf" to compute the chemical energy due to catalytic surface reactions (such as exothermic recombination reactions). Also now take into account that chemical energy in the total energy,

## Author(s)

Arnaud Borner (NASA Ames)

## Backward Compatibility

Backwards compatible, except that a new surface chemistry file needs to be used,.

## Implementation Notes

Added the heat of reaction to the surface reaction input file, added "echem" as an option to "compute surf", and added the chemical surface catalytic energy to "etot". Edited doc.

## Post Submission Checklist


aborner1:
I have merged current master into this and it's passing all the reg tests. It should be ready to merge. Can we take care of it soon before it needs more manual merging?

stanmoore1:
@aborner1 I asked @sjplimp to review.

aborner1:
@stanmoore1 @sjplimp Ditto here: Talked to Michael and he green-lighted this. Can we proceed with review and merge?

stanmoore1:
@aborner1 the changes here need to be ported to Kokkos, should have time to work on it this week.

aborner1:
@stanmoore1 Have you had a chance to look at the Kokkos version of this command?

stanmoore1:
@aborner1 I'm porting this PR to Kokkos now. Same as the other PR, if you have an example that would be helpful. Thanks

aborner1:
@stanmoore1 Yes, I can very easily do that for this one, super simple.

aborner1:
> @stanmoore1 Yes, I can very easily do that for this one, super simple.

done

stanmoore1:
Thanks @aborner1!

stanmoore1:
@aborner1 this is nearly ready to merge. One loose end I see is that the new feature only applies to the `prob` style surf reaction, `global` and `adsorb` are ignored. I think this should be mentioned in the docs.

aborner1:
> @aborner1 this is nearly ready to merge. One loose end I see is that the new feature only applies to the `prob` style surf reaction, `global` and `adsorb` are ignored. I think this should be mentioned in the docs.

Agreed. Are you able to add a quick mention while you are working on this?

stanmoore1:
Yes will do

aborner1:
> @aborner1 It also looks like this is going to force all SPARTA reaction files to have 2 coefficients instead of 1, correct? It seems like it may be better to have this optional in the reaction file, i.e. default the second coeff to 0.0. @sjplimp do you agree?

That is correct, and I am fine with defaulting to a 0.0 value, which would replicate the previous behavior.

PR #279, opened 2021-08-16
TCE chemistry enhancement (new option) + option to compute chemistry rates without performing reaction

aborner1:
## Purpose

New feature: adds a second option in the model to use the total energy of the participating particles (translational + rotational + vibrational), on top of the previous model that uses translational + a fraction of the rotational energy. Also adds a flag to only compute the chemical rates for reactions without performing them.

## Author(s)

Arnaud Borner (NASA Ames)
Mitchell Gosma (UIUC) - for polyatomics chemistry with discrete vibrational energy model 

## Backward Compatibility

Backwards compatible.

## Implementation Notes

The previous TCE model ("rDOF" from Bird remains the default). Adds a new second option that uses the total energy of the participating particles (translational + rotational + vibrational) and computes the participating degrees of freedom of those particles based on the temperature and vibrational model. The choice is made with the "partial_energy" yes or no keyword in "react_modify".

Also, adds a new keyword "compute_chem_rates", that, if specified as "yes" will compute the number of accepted chemical reactions without performing them, therefore enabling the computation of chemical rates (i.e. useful for a 0D simulation).

Added new doc for react_modify.

## Post Submission Checklist


stanmoore1:
@aborner1 Michael green-lighted this--I need to check if anything needs to be ported to Kokkos, then should be ready to merge.

stanmoore1:
@aborner1 do you an example for this that we can add to the repo? I'm porting this PR to Kokkos now.

aborner1:
@stanmoore1 Yes, I do. But also, I do now realize I want to port new changes to this. In particular, I want to remove the use of cell temperatures, and use collision temperatures (as Michael had suggested in the past). I have the changes all implemented, but it's on another branch, I will work on it ASAP. I would suggest you hold fire on Kokkos port in the meantime.

stanmoore1:
@aborner1 sounds good, thanks!

aborner1:
@stanmoore1 added example and updated files. I tried my best to update the kokkos files as well, but at least react_tce_kokkos.h/cpp need a bit more work, but hopefully only a few minutes. Otherwise, it is complete from my perspective.

stanmoore1:
Thank you @aborner1, I will take a look.

src/react_bird.cpp
```diff
@@ -248,13 +248,13 @@ void ReactBird::init()
 
     double c1 = MY_PIS*epsilon*r->coeff[2]/(2.0*sigma) *
       sqrt(mr/(2.0*update->boltz*tref)) *
-      pow(tref,1.0-omega)/pow(update->boltz,r->coeff[3]-1.0+omega) *
-      tgamma(z+2.5-omega) / MAX(1.0e-6,tgamma(z+r->coeff[3]+1.5));
+      pow(tref,1.0-omega)/pow(update->boltz,r->coeff[3]-1.0+omega); // *
+//      tgamma(z+2.5-omega) / MAX(1.0e-6,tgamma(z+r->coeff[3]+1.5));
```
stanmoore1, line 252: @aborner1 what is the purpose of this change? Seems unrelated to computing chemistry rates with performing reactions?

stanmoore1 (commented):
@aborner1 this PR breaks the ambipolar regression test. Can you explain the changes I pointed out?

src/react_bird.cpp
```diff
@@ -248,13 +248,13 @@ void ReactBird::init()
...
     double c2 = r->coeff[3] - 1.0 + omega;
 
     r->coeff[2] = c1;
-    r->coeff[3] = c2;
-    r->coeff[5] = z + 1.5 - omega;
+//    r->coeff[3] = c2;
```
stanmoore1, line 256: Same--seems unrelated to PR topic.

src/react_tce.cpp
```diff
@@ -73,22 +84,76 @@ int ReactTCE::attempt(Particle::OnePart *ip, Particle::OnePart *jp,
...
 
     switch (r->type) {
     case DISSOCIATION:
     case IONIZATION:
     case EXCHANGE:
       {
-        react_prob += r->coeff[2] *
-          pow(ecc-r->coeff[1],r->coeff[3]) *
-          pow(1.0-r->coeff[1]/ecc,r->coeff[5]);
+        react_prob += r->coeff[2] * tgamma(z+2.5-r->coeff[5]) / tgamma(z+r->coeff[3]+1.5) *
```
stanmoore1, line 154: Same.

src/react_bird.cpp
aborner1: "z" is now variable and needs to be computed on a per reaction basis in react_tce as opposed to a priori in react_bird, hence why it got moved there.

src/react_bird.cpp
aborner1: Same explanation as above. Moved to react_tce

src/react_bird.cpp
stanmoore1: I see. Any ideas why the `ambi` test would fail?

stanmoore1:
Here is the output from the ambi test:
```
13:15:15 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
13:15:15 dir = /var/jenkins/workspace/SPARTA_CPU_testing/dsmc_testing/examples/ambi
13:15:15 test = ambi
13:15:15 gold standard = log.archive.03Oct23.linux.mpi_4.ambi
13:15:19 1000
13:15:19 read 11 log entries
13:15:19 1000
13:15:19 read 11 log entries
13:15:19 ==== comparing log.mpi_4.ambi with log.archive.03Oct23.linux.mpi_4.ambi ====
13:15:19 Step     error  0.0 wrt norm  1000.0
13:15:19 Np       error 145.0 wrt norm 147746.0
13:15:19 Natt     error 1738.0 wrt norm 45316.0
13:15:19 Ncoll    error 406.0 wrt norm 20140.0
13:15:19 Nsreact           error  1.0
13:15:19 c_temp   error 158.08 wrt norm 180627.86
13:15:19 c_10[6]  error 77.0 wrt norm    37.0
13:15:19 c_10[7]  error  3.0 wrt norm     7.0
13:15:19 c_10[8]  error 206.0 wrt norm   115.0
13:15:19 c_10[9]  error 350.0 wrt norm   885.0
13:15:19 c_10[10]  error 12.0 wrt norm    27.0
13:15:19 c_10[11]           error  0.0
13:15:19 !!! test ambi FAILED
13:15:19 elapsed time = 4.4251229763 s for test ambi
```

src/react_bird.cpp
aborner1: Yes. I removed the  "MAX(1.0e-6,..." part, because I don't think it should be there.
If you add it back at line 154, the test will pass.

Now, as to why I did that.
This was a fix from many years ago, when we realized that the gamma function would overflow if its argument was negative. In my opinion, I would prefer that either a warning is issued to the user, or the code stops running, when the temperature exponent of a certain reaction is out of range (i.e., too negative), as opposite to silently run (which is the case now), the user not be made aware.
The O and N ionization reactions (last 2 in tce file) have temperature exponents that should not be allowed in the TCE model because they lead to negative gamma function arguments, so they need to be refitted to have a temperature exponent close to 0.

This is more philosophical than anything else.
So I think the short version is, if you want the reg test to pass, add the "MAX" part back in, but I still think the whole "temperature exponent" check should eventually be addressed, but I guess it should be its own PR.

stanmoore1:
@aborner1 do we need to mention in the docs that this only works for TCE, not QK?

aborner1:
@stanmoore1 - yes, good point

stanmoore1:
OK I'll also add an error check.

stanmoore1:
@sjplimp Waiting for regression testing to finish--I think this is now ready to review/merge.

stanmoore1 (approved):

stanmoore1:
@aborner1 I tweaked the code so no changes to collide are necessary for the "compute chem rates" option. However, one side effect is that the `Nreact` variable in the stats output now gives 0's when this option is enabled. Is this acceptable? I think it makes sense since no reactions are actually being performed?

aborner1:
@stanmoore1 Yes, that makes sense, and I have personally don't have a problem with that. So I think we can move forward with that.

stanmoore1:
@aborner1 thanks, waiting on final review from @sjplimp, this is ready to merge from my POV.

sjplimp (approved):

PR #281, opened 2021-09-02
Relax cell temp

zseckert:
## Purpose

This PR adds the ability to use per-cell temperatures to calculate relaxation collision numbers, rather than collision temperatures. Both computes and variables can be used, which should allow sufficient generality for users to decide exactly what values are used in the calculation.

## Backward Compatibility

The default behavior, when the new input syntax is omitted, matches the code paths of the old version. No changes to the outputs of old input decks should occur.

## Implementation Notes

Two new tests/examples have been added, one for the compute version and one for the variable version. Documentation has been added to the collide page.

This change has not yet been ported to the Kokkos version

This did require adding a temperature parameter to several existing methods in collide and collide_vss. This may have farther reaching consequences than I understand, but we can iterate to find something that works if so.

## Post Submission Checklist


zseckert:
@stanmoore1 I was hoping to circle back to this and get it merged to avoid any more conflicts developing and to preserve the functionality because Michael and I have made some recent use of it. I think the only big task remaining is also implementing it in the Kokkos version, however I have never worked with the Kokkos version. Before I dive into it, I wanted to ask about possible pitfalls. Are the sources similar enough that porting the diff from master to the Kokkos will be sufficient, or will be more involved?

stanmoore1:
Hi @zseckert I will take a look, I may be able to quickly port this to Kokkos easier than you can learn Kokkos, unless you want to.

zseckert:
@stanmoore1 I agree that it would likely be quicker for you to port it. After I'll take a look at the diffs and fulfill my own curiosity. Early next week I'll also take a look at the merge conflict in collide_vss.cpp.

aborner1:
> @stanmoore1 I was hoping to circle back to this and get it merged to avoid any more conflicts developing and to preserve the functionality because Michael and I have made some recent use of it. I think the only big task remaining is also implementing it in the Kokkos version, however I have never worked with the Kokkos version. Before I dive into it, I wanted to ask about possible pitfalls. Are the sources similar enough that porting the diff from master to the Kokkos will be sufficient, or will be more involved?

I have a related PR on collide_vss that should likely only be pushed once @zseckert 's is approved and merged, so I will wait
until that is approved and merged, but I also strongly support quick merging. For what it's worth, I looked at it and it all looks good to me. Obviously Kokkos version still needs modified.

stanmoore1:
OK I have to do something for LAMMPS today, but will work on this and the other SPARTA pull requests tomorrow or Monday. @zseckert @aborner1

stanmoore1:
@zseckert I looked at the code and I see how to do this for Kokkos. Will need a small refactor of the code though.

sjplimp:
@zseckert This is an interesting idea - thanks for submitting.  There are some coding changes that would be needed.  But let me ask a few conceptual Qs first.

a) Which computes produce something useful in this context?  I'm thinking not all of them produce a vector output, which is not checked for?
b) Likewise what are examples of variables that are useful?
c) Invoking a compute at this point in the timestep is not something we envisioned.   It's possible it creates problems in some scenarios.  I think this line you added to collde.cpp
if (!(modify->compute[icompute]->invoked_flag & INVOKED_PER_GRID)) {
        modify->compute[icompute]->invoked_flag |= INVOKED_PER_GRID;
is to unset the INVOKED_PER_GRID flag so the compute will be re-invoked later in the timestep if it is requested?
Isn't there a similar issue if a variable a compute?
d) For a large calculation (many particles) how much does invoking a per-grid compute every step slow down the calculation?
I.e. is this practical?

Steve

zseckert:
@sjplimp I have discussed this with Michael, and we have come to the conclusion, given your comments, that merging this functionality may not be necessary/worth our while. We have a manuscript under review that compares methods that use a cell averaged temperature to other methods that use only properties of the colliding particles, and (in our opinion) the later methods are superior. The goal of merging it was to maintain this functionality to reproduce the work of that paper, but given the conclusions, archiving it on a branch is also acceptable.

To respond to your comments directly:
a) and b) it is primarily intended to work with computes of temperatures and variables that are some weighted average of temperatures. There are many different models in the literature and conceivably a user could want to use something else, hence the generality, but these are the only uses I can think of at the moment.
c) I think your interpretation is correct. The functionality could work with a compute calculated at any point in the timestep, so a refactoring might alleviate unexpected behavior. Given the conclusion above, this reworking is likely not worth it at present.
d) In my simulations of 10M particles in anywhere from 27 to 3M cells, the added cost of the compute was moderate. For many particles per cell (~100k), the compute added 10-20% (the only apples-to-apples comparison in my existing data). As the number of particles per cell dropped, the cost increased significantly to roughly tripling the simulation wall time at 6 particles per cell. However, some fraction of this could well be moving particles between cells, and not the compute. I did not run the cell based temperatures for discrete vibrational mode, but those simulations were significantly slower as well, and I would expect the compute would be a smaller fraction of the runtime in those cases.

sjplimp:
@zseckert Thanks for the detailed explanations.  Given the results in your pending paper, I agree it is probably not so useful to include this option in the public code.  So we can close the pull request.  I believe that will still leave the current state of the PR as well as all these useful discussion notes stored in the GitHub repo, so that we can revisit things at some point if that becomes useful.  Thanks again for working on this - and your paper sounds interesting.

@stanmoore1 Can you verify what I just said, and close the PR ?

PR #284, opened 2021-09-22
Add mars/venus data

aborner1:
New species, vss and tce file. Also edit co2.species.vib (since CO2 has 4 modes), and remove co2.vss and co2.species.

## Purpose

Add new files to /data folder. No code changes.

## Author(s)

Arnaud Borner (NASA Ames)
Krishnan Swaminathan-Gopalan (NASA Ames)

## Backward Compatibility

Fully compatible.


aborner1:
@stanmoore1 I have restored the CO2 species and vib files, as well as edited the species.vib file, /data/. I have also edited the species and species.vib in the vibrate example to better reflect reality (CO2 has 4 vib modes and not 3). Therefore the vibrate reg test will fail because the results are different. I think the reg test should be updated to this example.
The mars files are still present in /data/.
Ready to merge is you are ok with slightly changing the example/test.

stanmoore1 (approved):

PR #291, opened 2021-10-20
Fix issue with reaction index not being passed to tally computes

sjplimp:
## Purpose

The recent PR to add the surf_react adsorb command broke the way compute react/surf and compute react/boundary work
using the index of the performed reaction.  This was due to a change in the way the surf react classes interact with the calling surf_collide classes, to enable either class to set new velocities for the reacted particles.  This change uses 2 returned values to support both operations.

## Author(s)

Steve

## Backward Compatibility

Should fix a bug.

## Implementation Notes

## Post Submission Checklist

### DIFF
--- a/src/cut2d.cpp
+++ b/src/cut2d.cpp
@@ -311,7 +311,7 @@ int Cut2d::split(cellint id_caller, double *lo_caller, double *hi_caller,
 
   while (1) {
     build_clines();
-    
+
 #ifdef VERBOSE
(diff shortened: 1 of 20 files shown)


sjplimp:
@stanmoore1 If this passes regression tests, particulary for the new surf react/adsorb input scripts, then the version patch release PR should also be ready to go.

@timteichmann Good catch on flagging this - see if this addresses your concerns?

PR #316, opened 2022-01-07
Update restart file to include all global options

### LINKED ISSUE #162: Cellmax is not saved in the restart file if changed by user
## Summary

Cellmax variable is not saved and cannot be read by the read_restart function if user changes it.

## Type of Issue

Bug Report

Latest version downloaded from Github

The code should be able to restart if the user sets the global cellmax variable to a value different than 100. 

When you try to restart the code with a restart script when it reads the restart file the it pops an error message saying that the global cellmax is not set although it is set in the input script.

I checked the read_restart, write_restart and i saw that the CELLMAX value is not taken into account. Tried to add to enumerate the CELLMAX changing some other lines (in order to write and read the new variable) but then i get error messages from the processes.

Make a surface file with more than 100 surfaces per cell and then set the global cellmax variable to something else than 100. When you try to restart you get an error message that the cellmax value is not set.

stanmoore1:
## Purpose

Add missing `surfs`, `surfgrid`, `splitmax`, `surftally`, `particle/reorder`, and `mem/limit` global options to restart file.
Closes #162.

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

This breaks restart file backwards compatibility.


sjplimp (approved):

sjplimp:
@stanmoore1 Before merging, can you make sure these changes are consistent with the read_restart doc page?  It mentions
several of the global settings.  Want to make sure we are not breaking some condition it mentions, and that that doc page is clear that all the global settings are in the restart file.

stanmoore1:
> @stanmoore1 Before merging, can you make sure these changes are consistent with the read_restart doc page? It mentions several of the global settings. Want to make sure we are not breaking some condition it mentions, and that that doc page is clear that all the global settings are in the restart file.

Yeah will do.

stanmoore1:
@sjplimp I checked the read_restart doc page, it says all global commands are stored, which is now true, so everything should be consistent.

PR #321, opened 2022-01-19
Fix surface reaction reference in variable

timteichmann:
## Purpose

Fixes a bug in variable. Currently the lookup for the surface reaction model falsely uses the function for the collision model. Thus, the lookup either fails or returns a wrong reference.

## Author(s)

Tim Teichmann, KIT

## Backward Compatibility

Functionality was broken before.

## Implementation Notes

Trivial: use correct lookup function


stanmoore1:
@timteichmann good catch!

stanmoore1 (approved):

PR #324, opened 2022-01-31
Add no slip option to surf_collide_specular

stanmoore1:
## Purpose

Add `noslip` option to `surf_collide_specular`.  In this, all velocity components are reflected off the surface (instead of just the normal component), which creates a no-slip condition on the surface.

## Author(s)

Michael Gallis (SNL), Stan Moore (SNL)

## Backward Compatibility

Yes

### DIFF
--- a/doc/surf_collide.txt
+++ b/doc/surf_collide.txt
@@ -15,8 +15,8 @@ surf_collide ID style args keyword values ... :pre
 ID = user-assigned name for the surface collision model :ulb,l
 style = {specular} or {diffuse} or {cll} or {impulsive} or {td} or {piston} or {transparent} or {vanish} or {specular/kk} or {diffuse/kk} or {piston/kk} or {vanish/kk} :l
 args = arguments for specific style :l
-  {specular} or {specular/kk} args = adiabatic (optional)
-    adiabatic = reflect all velocity components off surface (not just normal component)
+  {specular} or {specular/kk} args = noslip (optional)
+    noslip = reflect all velocity components off surface (not just normal component)
   {diffuse} or {diffuse/kk} args = Tsurf acc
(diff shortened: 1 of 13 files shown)


src/surf_collide_specular.cpp
```diff
@@ -27,7 +27,19 @@ using namespace SPARTA_NS;
...
+  // optional args
+
+  adiabatic_flag = 0;
+
+  int iarg = 2;
+  while (iarg < narg) {
+    if (strcmp(arg[iarg],"adiabatic") == 0) {
+      adiabatic_flag = 1;
+      iarg += 1;
+    }
```
sjplimp, line 41: this needs an error test in case adiabatic is misspelled or a wrong keyword is used

stanmoore1: Good point, will fix

doc/surf_collide.txt
```diff
@@ -115,7 +116,10 @@ The {specular} style computes a simple specular reflection model.  It
 requires no arguments.  Specular reflection means that a particle
 reflects off a surface element with its incident velocity vector
 reversed with respect to the outward normal of the surface element.
-The particle's speed is unchanged.
+The particle's speed is unchanged. If the optional {adiabatic} keyword
```
sjplimp, line 119: Specular reflection means that a particle bounce off a surface element
reverses only the component of its velocity normal to the surface.  If the optional {adiabatic} keyword
is used, the particle bounce flips the sign of all 3 xyz components of the particle's incident
velocity, so that it now moves in the opposite direction.  In either case, the particle's speed is unchanged.

sjplimp (changes requested):
One change in code, one change in doc page suggested.

PR #329, opened 2022-02-16
Add surf_collide adiabatic model with isotropic scattering

timteichmann:
## Purpose

This adds a new surface collision model to model adiabatic surfaces. This model was proposed by Mohammadzadeh, Rana, and Struchtrup  [1]. Basically, particles that hit the surface are isotropcially scattered (similar to a diffuse reflection) but their velocity magnitude is conserved. This means that the kinetic energy of the particle remains unchanged and no energy is transferred from or to the surface (adiabatic). Because this is done per particle no iterative procedure is necessary to determine the surface temperature. Note, that due to the isotropic scattering momentum is not conserved, thus still leading to a non-slip condition on the wall.

## Author(s)

Tim Teichmann (ITEP, KIT)

## Backward Compatibility

Optional feature

## Implementation Notes

As discussed by mail this is the first version upon which we can iterate.

Since I have no exp with KOKKOS I have not included a dedicated KOKKOS version for now. I have added two examples in the surf_collide dir, especially the beam example illustrates the isotropic scattering when compared with the specular case.

If you like I can add another example, best would be something like a box filled with particles and adiabatic walls that demonstrates that energy is conserved.

## Post Submission Checklist

[1] A. Mohammadzadeh, A. Rana, and H. Struchtrup, "DSMC and R13 modeling of the adiabatic surface", International Journal of Thermal Sciences, vol. 101, pp. 9–23, March (2016), doi: 10.1016/j.ijthermalsci.2015.10.007.

### DIFF
--- a/doc/surf_collide.html
+++ b/doc/surf_collide.html
@@ -17,7 +17,7 @@
 </PRE>
 <UL><LI>ID = user-assigned name for the surface collision model 
 
-<LI>style = <I>specular</I> or <I>diffuse</I> or <I>cll</I> or <I>impulsive</I> or <I>td</I> or <I>piston</I> or <I>transparent</I> or <I>vanish</I> or <I>specular/kk</I> or <I>diffuse/kk</I> or <I>piston/kk</I> or <I>vanish/kk</I> 
+<LI>style = <I>specular</I> or <I>diffuse</I> or <I>cll</I> or <I>adiabatic</I> or <I>impulsive</I> or <I>td</I> or <I>piston</I> or <I>transparent</I> or <I>vanish</I> or <I>specular/kk</I> or <I>diffuse/kk</I> or <I>piston/kk</I> or <I>vanish/kk</I> 
 
 <LI>args = arguments for specific style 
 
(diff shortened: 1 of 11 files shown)


timteichmann:
We could also merge the specular option in here and require an additional argument {isotropic,specular}. Implementation would be straightforward by including an if clause that decides whether to call `scatter_isotropic(p,n)` or `scatter_specular(p,n)`. Maybe this would be less confusing than having two implementations?
Let me know if you want me to include the necessary changes.

stanmoore1:
Thanks @timteichmann!

stanmoore1:
Hi @timteichmann , I talked to @sjplimp and we decided to rename the `adiabatic` keyword in Specular to `noslip`, and will keep your isotropic adiabatic scattering as `surf_collide_adiabatic`. One question: did you mean to allow your new method to work with `surf adsorb`? We saw you added the `wrapper` method and the `allowreact` flag is not set to `0` in your method.

timteichmann:
Hi @stanmoore1, alright that's also a good solution!
Regarding your question on reactions: I think there is no reason to not allow them. You may have seen in the comments to the code that I mentioned that reactions might not conserve energy though. The SPARTA version in which I had implemented this locally was a bit older (Feb21) and thus `react_adsorb` was not yet present. However, when porting I saw no programmatic reason not to support it? I see it then in the responsibility of the user to make an informed decision whether a specific combination is physically reasonable or not.

stanmoore1:
OK I'll update the doc page.

timteichmann:
Thanks!

stanmoore1 (approved):

PR #331, opened 2022-02-22
Remove superfluous include in variable.cpp

timteichmann:
## Purpose

I think one include is enough ;-)

## Author(s)

Tim Teichmann (ITEP,KIT)

## Backward Compatibility

Yes, compiler fixed this for us anyways.

## Implementation Notes

Trivial


stanmoore1:
@timteichmann good catch!

stanmoore1 (approved):

PR #355, opened 2022-08-22
Add Py3 support

stefanfechter:
This commit adds support for Python3 to the Sparta Python interface as
well as it updates the Sparta tools for the usage of Python3. Mainly
covers adaptions to the print commands etc.

## Purpose

This adds support for Python3 to the Sparta python files as well as the Sparta Python interface.

## Author(s)

Stefan Fechter (DLR)

## Backward Compatibility

No, works with Python2 and Python3 versions.

## Implementation Notes

Python files transformed using the Python tool 2to3.

## Post Submission Checklist


stanmoore1 (approved):
Looks good to me.

python/examples/demo.py
```diff
@@ -27,25 +27,25 @@
 
 # test out various library functions after running in.demo
 
-spa.file("in.demo")
+spa.file(b"in.demo")
```
stanmoore1, line 30: @stefanfechter just noticed there are some extraneous ` b` characters, seems to only be in the demo.py file.

stanmoore1 (changes requested):

python/examples/demo.py
```diff
@@ -29,23 +29,23 @@
 
 spa.file("in.demo")
```
stefanfechter, line 30: The "b" before the string is needed here due to the different string treatment in Python3; otherwise it does not work with Python3.

stefanfechter: for more information, see e.g.:
https://stackoverflow.com/questions/2592764/what-does-a-b-prefix-before-a-python-string-mean

In Python3 the string syntax has changed, but with the included __future__ statements it is backward compatible to Python2 > 2.6.

stanmoore1: >The "b" before the string is needed here due to the different string treatment in Python3; otherwise it does not work with Python3.

I see, sorry about that. I'm going to defer to @sjplimp on this pull request.

stanmoore1 (approved):

stefanfechter:
@sjplimp Yes it should work with Python2 and Python3 (at least that claims the program 2to3 together with the future imports for Py2). On my system I've checked it with Py2 as well, but I have all Py3 packages installed. Feel free to double-check it on your system as well.

sjplimp (approved):

zseckert:
I just tested this on my rhel8 laptop. An important note is that by default rhel8 does not have a `python` executable on the PATH, and no python 2 installed at all, only `python3`. To get the tests to run I had to change `cmake/common/test/sparta_test_utils.cmake`.  At the very top I added `find_package( Python REQUIRED)` then in the two places where `python` was called directly below I changed to `Python::Interpreter`.

I am seeing some test failures (collideInterspecies, thermostat, adjust_temp) with diffs on the order of 1-2%. I have seen some different behavior in other codes due to the compilers on this machine, so these are likely not due to the python changes.

stefanfechter:
Thanks @zseckert  for the testing the patchset. Until now the Python part is not tested using the CMake build system. I followed the instructions in the manual and that way it worked.

I can confirm the test failures on my Ubuntu 20.04 system as described (collideInterspecies, thermostat, adjust_temp). Seems to be related to newer operating systems.

zseckert:
@stanmoore1 I think so. Using find_package should be more robust than depending on the PATH within CMake.

stanmoore1:
@zseckert does this look good to you?

zseckert:
@stanmoore1 Yes, this looks fine to me.

PR #368, opened 2022-10-06
Fix bug regarding transparent surfaces in grid_surf

timteichmann:
## Purpose

Discovered this while I was tracking something different. I have not seen this blow up, but the code makes no sense if the conditionals do not include the `break` as it will always only inspect the first element in the `for`-loop in the current implementation? If I am wrong feel free to just close this.

## Author(s)

Tim Teichmann (KIT)

## Backward Compatibility

I think it should be compatible, but not 100% sure.

## Post Submission Checklist

- [] One or more example input decks are included


stanmoore1 (approved):

stanmoore1:
@timteichmann good catch, this does look like a bug.

PR #382, opened 2023-01-16
Update Python interface to Python3

sjplimp:
## Purpose

Change the python/sparta.py file to be Python 3 compatible, which includes its use of ctypes to interface to the SPARTA library.

## Author(s)

Steve

## Backward Compatibility

N/A so long as you use the Python interface from Python 3 instead of Python 2

## Implementation Notes

N/A

## Post Submission Checklist

### DIFF
--- a/python/sparta.py
+++ b/python/sparta.py
@@ -29,7 +29,7 @@ class sparta:
     except:
       type,value,tb = sys.exc_info()
       traceback.print_exception(type,value,tb)
-      raise OSError("Could not load SPARTA dynamic library")
+      raise Exception("Could not load SPARTA dynamic library")
 
     # create an instance of SPARTA
     # don't know how to pass an MPI communicator from PyPar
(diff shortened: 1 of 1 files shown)


sjplimp:
@stefanfechter Hi Stefan - I just submitted this PR to fix an issue I was having with the Python interface, with this test:

```
% cd src; make mode=shlib mpi     # creates the shared SPARTA lib for Python to load
% cd examples/collide
% python
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from sparta import sparta
>>> spa = sparta()
SPARTA (18 Jul 2022)
Running on 1 MPI task(s)
>>> spa.file("in.collide")
ERROR on proc 0: Cannot open input script i (../input.cpp:227)
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------
```

This is with the version in current master.  My edits to sparta.py in this PR, let the in.collide script run as expected.
I believe this was an issue with the Python ctypes interface in Python 3 (used to work in Python 2).

Looking back at your PR in Aug 2022 (which is what is in current master), I am a bit confused.
Do you agree that this new version is still correct with regards to what you changed?
Also I am not sure if my new version works with either Py3 or Py2 - maybe that is not so important anymore.

sjplimp:
@stefanfechter I'll also note that the ctypes issue was with passing a Python3 string to a C function in the SPARTA lib interaface.
The C function was just seeing a single char (not the whole string), hence the error.  I believe in Py3, the string needs
to be byte-encoded with utf-8, before being passed a c_char_p thru ctypes.

stefanfechter:
@sjplimp Hi Steve,
yes that is due to a change in the string handling in Python3. I adapted the example in "python/demo/demo.py" for the usage.

Another option would be to add a "b" into the calling of the Python interface (this option I choose at the end) that tells the interpreter to use a byte-compiled version of the string (the same that you did with multiple changes on the Python interface).
>     In [1]: from sparta import sparta
>     In [2]: spa = sparta()
>     SPARTA (18 Jul 2022)
>     Running on 1 MPI task(s)
>     In [3]: spa.file(b"in.collide")

In my opinion this does not change the backward compatibility with older Python2 versions.

sjplimp:
@stefanfechter thanks, but if you are doing this with a string variable (file or name in this code) instead of a string constant, do you have to do it the way this PR is doing it?   The code I assed seems a bit clunky

python/sparta.py
```diff
@@ -56,33 +56,36 @@ def close(self):
...
+    b_file = file.encode('utf-8')
+    self.lib.sparta_file(self.spa,c_char_p(b_file))
 
   def command(self,cmd):
-    self.lib.sparta_command(self.spa,cmd)
+    b_cmd = cmd.encode('utf-8')
+    self.lib.sparta_command(self.spa,c_char_p(b_cmd))
 
   def extract_global(self,name,type):
+    b_name = name.encode('utf-8')
```
stefanfechter, line 67: That is what the u''-String prefix does in your Python program. The main question is whether we want to let the user do that in the scripts or do that in an interface layer.

python/sparta.py
```diff
@@ -56,33 +56,36 @@ def close(self):
     self.spa = None
 
   def file(self,file):
-    self.lib.sparta_file(self.spa,file)
+    b_file = file.encode('utf-8')
+    self.lib.sparta_file(self.spa,c_char_p(b_file))
```
stefanfechter, line 60: The `c_char_p` is not needed on my end.

stefanfechter (commented):
@sjplimp 
Thanks for digging into this Python3 related stuff. The main question, I've raised in the code discussion. In my opinion the `c_char_p` is not needed and the main thing is, if we expect byte-strings from the users.

python/sparta.py
sjplimp: ah, I see what your earlier comment meant now.  I don't think we should assume or require the user to pass in a byte-string.

python/sparta.py
sjplimp: thanks @stefanfechter - I will try this w/out on my box

sjplimp: @stefanfechter you're right - the c_chap_p() cast is not needed - so I removed those, leaving the encoding function there - thanks!

sjplimp:
I think this is ready to merge

stanmoore1 (approved):

PR #385, opened 2023-02-02
variable time stepping

akstagg:
## Purpose

This merge adds a fix to enable calculation of a global, variable time step.

## Author(s)

Alan Stagg (SNL)

## Backward Compatibility

The addition to the restart files of the current time and timestep breaks backward compatibility in the reading of restart files created before this merge.

## Implementation Notes

The primary logic for the global timestep calculation is located in fix_dt.cpp, fix_dt_kokkos.cpp, and associated headers.  The implementation follows the discussion in the papers referenced below but does not include spatial timestep adaptivity.  The global timestep is calculated as a function of the local cell timestep values, which are constructed from cell mean collision and transit times.

## Post Submission Checklist

doi:10.1063/1.3067865
doi:10.1016/j.jcp.2009.03.021


stanmoore1:
@akstagg thanks, will take a look

doc/fix_dt.txt
```diff
@@ -0,0 +1,85 @@
...
+[Example:]
+
+stats_style     step time dt f_DT cpu np :pre
+
+:line
+
+[Restart, output info:]
+
+The global time and the current global timestep are written to "binary restart files"_restart.html.  This breaks backward compatibility
+with older restart files, even if this fix is not invoked.
```
stanmoore1, line 70: @akstagg can you remove this note about backward compatibility in the docs--we will post this in the patch release notes.

src/KOKKOS/collide_vss_kokkos.cpp
```diff
@@ -626,7 +626,7 @@ void CollideVSSKokkos::operator()(TagCollideCollisionsOne< NEARCP, ATOMIC_REDUCT
   // attempt = exact collision attempt count for a pair of groups
   // nattempt = rounded attempt with RN
 
-  const double attempt = attempt_collision_kokkos(icell,np,volume,rand_gen);
+  const double attempt = attempt_collision_kokkos(icell,np,volume,dt,rand_gen);
```
stanmoore1, line 629: Maybe I am misunderstanding, but it looks like these changes to add `dt` into the functions is not needed?

src/collide.cpp
```diff
@@ -498,7 +499,7 @@ template < int NEARCP > void Collide::collisions_one()
     // nattempt = rounded attempt with RN
     // if no attempts, continue to next grid cell
 
-    attempt = attempt_collision(icell,np,volume);
+    attempt = attempt_collision(icell,np,volume,dt);
```
stanmoore1, line 502: Ditto here, not sure why `dt` needs to be part of the function arguments?

src/run.cpp
```diff
@@ -11,7 +11,7 @@
 
    See the README file in the top-level SPARTA directory.
 ------------------------------------------------------------------------- */
-
+#include <iostream>
```
stanmoore1, line 14: Is this leftover from debugging? We normally don't use iostream in SPARTA.

src/update.cpp
```diff
@@ -11,7 +11,7 @@
 
    See the README file in the top-level SPARTA directory.
 ------------------------------------------------------------------------- */
-
+#include <iostream>
```
stanmoore1, line 14: Leftover debugging?

stanmoore1 (changes requested):
@akstagg this looks really good and close. I put a few minor change requests in the review.

stanmoore1 (approved):
I approve. @sjplimp can you take a look?

stanmoore1 (approved):

PR #397, opened 2023-02-23
Fix bug with Kokkos bitmasks

stanmoore1:
## Purpose

Fix bug with Kokkos bitmasks, where data wasn't always being transferred between GPU and CPU correctly. This does not affect CPUs or GPUs when UVM is enabled. It manifested as a crash in the problem input if grid adaptation was enabled with the error: `Particle being sent to self proc on step 1`.

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

Yes


stanmoore1:
GPU regression tests passed.

PR #401, opened 2023-03-13
Add realloc_kokkos method, reduce Kokkos memory use

stanmoore1:
## Purpose

In order to reduce Kokkos memory use, it is necessary to free the memory before realloc'ing. This PR adds a helper function to do this, copied from LAMMPS. Also try to reduce memory footprint of the Kokkos version by removing unused views, etc.

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

Yes


stanmoore1:
Regression tests passed.

PR #412, opened 2023-05-16
Normfactor computation consistent with surfaces for axisymmetric case

knstmrd:
## Purpose

This small change adds additional factors to compute the surface area (that is used for flux scaling) of the box boundary in the axisymmetric case. This way this is consistent with computations done in `compute_surf.cpp` and `surf.cpp` (see `surf->axi_line_size`).  This uses the same scaling for YLO, although that surface/boundary has an area of 0 (due to being on the axis of rotation).

## Author(s)

Georgii Oblapenko, DLR

## Backward Compatibility

No issues with tests; no changes to existing input files needed. May cause inconsistency in output in computations that used the box boundary flux for the YHI boundary in an axisymmetric setup.

## Post Submission Checklist

### DIFF
--- a/src/compute_boundary.cpp
+++ b/src/compute_boundary.cpp
@@ -101,6 +101,11 @@ void ComputeBoundary::init()
   if (domain->dimension == 2) {
     normflux[XLO] = normflux[XHI] = domain->yprd * nfactor;
     normflux[YLO] = normflux[YHI] = domain->xprd * nfactor;
+
+    if (domain->axisymmetric) {
+      normflux[YLO] *= 2 * M_PI * domain->yprd;
+      normflux[YHI] *= 2 * M_PI * domain->yprd;
+    }
   } else if (domain->dimension == 3) {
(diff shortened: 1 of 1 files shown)


sjplimp:
@knstmrd good catch - but I'd like to make the code more clear for the axisymmetric case.

Please do something like this so there are 3 distinct cases:
if (dim == 2) {
  if (!axisym) {
     // existing 2d settings
  } else {
    nf[YLO} = nf[YHI] = 2.0 * PI * xprd * yprd;
 } else if (dim == 3) {
  // existing 3d settings
}
And add a comment that nf[YLO] is actually 0.0 in the axi case, but is used in the denom when the numerator also = 0.0

knstmrd:
@sjplimp I realized I've messed up the branching on my fork a bit, so I made a new branch with the updated version of the axisymmetric boundary normfactor; should I close this PR then and open a new one?

knstmrd:
Opened new PR: https://github.com/sparta/sparta/pull/421, closing this one.

PR #419, opened 2023-06-16
Fix random and proc style balance_grid input parse

knstmrd:
## Purpose

Fixes an issue with parsing the `balance_grid` command with `proc` or `random` style load-balancing options.

## Author(s)

Georgii Oblapenko, DLR

## Backward Compatibility

No issues with backward compatibility unless someone was relying on some internal tests breaking with balance_grid random or proc.

## Implementation Notes

Previous attempts to run with `balance_grid proc` or `balance_grid random` led to the following error: `ERROR: Illegal balance_grid command (../balance_grid.cpp:151)`, as the number of parsed string arguments was being erroneously set to 0.

## Post Submission Checklist


stanmoore1:
@knstmrd nice catch!

PR #421, opened 2023-06-21
Normfactor computation consistent with surfaces for axisymm case

knstmrd:
## Purpose

This small change adds additional factors to compute the surface area (that is used for flux scaling) of the box boundary in the axisymmetric case. This way this is consistent with computations done in `compute_surf.cpp` and `surf.cpp` (see `surf->axi_line_size`). This uses the same scaling for YLO, although that surface/boundary has an area of 0 (due to being on the axis of rotation).

## Author(s)

Georgii Oblapenko, DLR

## Backward Compatibility

No issues with tests; no changes to existing input files needed. May cause inconsistency in output in computations that used the box boundary flux for the YHI boundary in an axisymmetric setup.

## Implementation Notes

## Post Submission Checklist

### DIFF
--- a/src/compute_boundary.cpp
+++ b/src/compute_boundary.cpp
@@ -24,8 +24,10 @@
 #include "math_extra.h"
 #include "memory.h"
 #include "error.h"
+#include "math_const.h"
 
 using namespace SPARTA_NS;
+using namespace MathConst;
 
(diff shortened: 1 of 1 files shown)


stanmoore1 (approved):
Looks good, no changes needed for Kokkos.

Issue #377, opened 2022-12-01
grid2paraview.py parse failure for valid create_box statement


tjotaha:
## Summary

grid2paraview.py parsing fails for this valid create_box statement.

create_box  1.75 8.25 5 10 -0.5 0.5

## Type of Issue

Bug Report

## SPARTA Version (Bug Report)

Development

## Expected Behavior (Bug Report)

grid2paraview.py parse should pass for the input

create_box  1.75 8.25 5 10 -0.5 0.5

## Actual Behavior (Bug Report)

grid2paraview.py parse fails with following output.

Error reading SPARTA grid description file
create_box specification is invalid:  create_box		1.75 8.25 5 10 -0.5 0.5

## Steps to Reproduce (Bug Report)

Run grid2paraview.py with the following grid.txt file.

dimension		2
create_box		1.75 8.25 5 10 -0.5 0.5
create_grid		750 500 1

knstmrd:
Stumbled on something similar myself.
Found the issue: lines 1034-1036 of paraview.py compare the box bounds without converting them to float from string.
I can open a pull request with a fix later on.

stanmoore1:
@knstmrd good catch, thanks!

PR #422, opened 2023-06-23
grid dimensions parsing fix

knstmrd:
## Purpose

Fixes an bug in the parsing of grid dimension in the Paraview conversion script. It was comparing strings instead of the actual float values, which would sometimes lead to errors depending on the grid sizing (see https://github.com/sparta/sparta/issues/377)

## Author(s)

Georgii Oblapenko

## Backward Compatibility

No changes in backward compatibility.

## Implementation Notes

Added float conversion in the grid dimensions check.

## Post Submission Checklist


stanmoore1 (approved):
@knstmrd nice catch, thank you!

Issue #429, opened 2023-07-31
Development of improved subsonic boundary condition and mass flow rate boundary condition


syamalpraneeth:
# Summary
We (@Cosimo89, @hieu325, @vimalrajv, and @syamalpraneeth) wish to make an improved implementation of the subsonic pressure in/out flow conditions---based on the existing fix/emit_surf in the SPARTA release and the work of Farbar and Boyd  [Farbar2014]. We explain the development we have made so far and challenges faced, and seek your help in successfully improving the boundary conditions.
# Type of Issue
Improving existing subsonic and mass flow rate boundary conditions.
# Detailed Description
We observed the following issues that motivated improvements to the official SPARTA (referred to as vanilla) code for our internal flow problems, as mentioned before in a previous issue thread: https://github.com/sparta/sparta/issues/413#issuecomment-1600266464:
* The fix/emit_surf_subsonic instruction is inaccurate in enforcing the target pressure at the inlet and outlet surfaces.
* We require a mass flow rate boundary condition at the inlet. Hence, we modified the subsonic pressure inlet to enforce a constant mass flow rate. This approach is based on  Farbar and Boyd  [Farbar2014], and summarized in Lei et al. [Lei2017].
The issue is organized as follows: In Section 1 we present the test case geometry and specifications, In Sections 2 and 3 we discuss the implementation of the pressure and mass flow rate boundary conditions, respectively. In Section 4 we report the current known bugs and remaining issues.  
# SPARTA Version and compilation details
* Based on Sparta 18Jul2022 release.
* Compiler toolchain is HPC-X v2.13.0. Compiled with KOKKOS: `cd sparta/src; make yes-kokkos; make kokkos_mpi_only`. 
* The `src/MAKE/Makefile.kokkos_mpi_only` file modified on line 32: `SPARTA_INC = -DSPARTA_GZIP –DSPARTA_BIGBIG`.

# Expected Behavior
## Section 1: Test geometry and model 
The model geometry is a rectangular Poiseuille flow of size LxHxW=0.01 m x 0.01 m x 0.0004 m. The sketch of the geometry, and the triangulated mesh are shown in Fig. 1. The surface mesh is converted from a .stl file to a SPARTA surface file using the stl2surf.py script. The surface file and model files (.species and .vss files) are made available in dedicated test folders (mentioned in later sections) with all the files required to reproduce the case. The flow is simulated using Ar gas atoms and the variable hard spheres (VHS) model.
[...]

PR #431, opened 2023-08-01
Do not emit particles in cells that are outside of the flow volume

timteichmann:
## Purpose

This fixes the emission of particles in cells that are outside of the flow volume. This fixes a problem for some case setups that feature internal flows (i.e. inside the geometry, e.g. in a pipe), where particles would be emitted twice for an inflow surface.

I suspect that this will also fix the bug mentioned in #429.

## Author(s)

Tim Teichmann (KIT, ITEP)

## Backward Compatibility

I have had this patch downstream for several years and have not found any issues.

## Post Submission Checklist


stanmoore1 (approved):

stanmoore1:
@timteichmann thanks, good catch!

PR #396, opened 2023-02-21
Port global and prob surface reactions to Kokkos

stanmoore1:
## Purpose

Port `global` and `prob` surface chemistry styles to Kokkos

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

The Kokkos package options `collide/retry` and `collide/extra` were renamed to `react/retry` and `react/extra`.

## Implementation Notes

I tested on both CPU and GPU.

PR #437, opened 2023-09-27
Fix Kokkos performance regression

stanmoore1:
## Purpose

Fix slowdown on NVIDIA GPUs from #396, reported by Matt Bettencourt (NVIDIA)

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

The Kokkos package `reduction` option was removed and the previous defaults will always be used now. This helps speed up compile time for `update_kokkos.cpp` which has many templates.


bathmatt:
Hi, 
I can confirm this removes the performance issue introduced, and even improves it slightly.  Not sure on approval process for SPARTA but for me, this is approved to fix the issue I noted.

stanmoore1:
@bathmatt thanks. We had to migrate our internal Jenkins tests from one server to another and it messed up some of the regression tests, which is causing the delay here. Should get it merged soon.

PR #446, opened 2024-01-02
Implemented electronic excitation

zseckert:
Added custom particle data and new input syntax to support arbitrary electronic states for each chemical species. The relaxation supports per-state and per-species relaxation collision numbers and allowed/forbidden transitions controlled by spin degeneracy.

## Purpose

This adds a, hopefully general, ability to simulate arbitrary electronic states with state-specific and species-specific relaxation collision numbers to SPARTA. There is also the ability to specify spin-degeneracy of the states separate from the total degeneracy which allows a rudimentary handling of allowed and forbidden transitions

## Author(s)

Primarily implemented/tested by myself, with input from Michael Gallis, and software design pointers from Steve Plimpton and Stan Moore

## Backward Compatibility

No breaking changes

## Implementation Notes

The new particle data uses the custom particle data, so no overhead is paid by those not using the feature. The output fixes/computes were expanded to interfaces with the new data as needed. The actual relaxation process was implemented in collide_vss.cpp which is fairly similar to discrete vibrational relaxation, albeit generalized. I have a suite of about half a dozen verification problems I ran as I was building the capability to ensure a quasi-0D simulation reached the expected analytic value. I can share these with anyone interested (they are too long running to include as examples). The included example is a version of them that touches all the various features/code paths, and if run out much longer (roughly 40k timesteps) will eventually reach the expected equilibrium.

As of yet, the KOKKOS implementation is untouched.

## Post Submission Checklist


zseckert:
I think I have fixed all the issues. I ran the included examples 'relax_electronic' out to 100k timesteps and it converged to the equilibrium as expected, and the Kokkos version now compiles, although none of the relaxation/parsing logic is implemented there.

stanmoore1:
@zseckert thanks I will port this to Kokkos

src/create_particles.cpp
```diff
@@ -379,6 +379,7 @@ void CreateParticles::create_single()
   double temp_thermal = particle->mixture[imix]->temp_thermal;
   double temp_rot = particle->mixture[imix]->temp_rot;
   double temp_vib = particle->mixture[imix]->temp_vib;
+  double temp_elec = particle->mixture[imix]->temp_elec;
```
stanmoore1, line 382: Unless I missed something, none of these new variables are used in the code (only defined)? Please remove or update. Similar for other files: `fix_emit_face.cpp`, `fix_emit_face_file.cpp`, and `fix_emit_surf.cpp`

src/particle.h
```diff
@@ -26,6 +26,8 @@ class Particle : protected Pointers {
   int sorted;               // 1 if particles are sorted by grid cell
 
   enum{MAXVIBMODE=4};       // increase value if species need more vib modes
+  enum{MAXELECSTATE=12};      // increase value if species need more electronic states
+  enum{MAXSPECIES=16};      // increase value if more species are needed for species-specific electronic relaxation rates
```
stanmoore1, line 30: `MAXSPECIES` name is confusing to use inside the `Species` struct, since the `species` vector will be `nspecies` long vector of the `Species` struct, so that seems like `nspecies` elements inside `nspecies` vector.

src/particle.h
```diff
@@ -37,11 +39,17 @@ class Particle : protected Pointers {
     double rottemp[3];      // rotational temperature(s)
     double vibtemp[MAXVIBMODE];   // vibrational temperature(s)
     double vibrel[MAXVIBMODE];    // inverse vibrational relaxation number(s)
+    double electemp[MAXELECSTATE];  // electronic temperatures for each electronic state
+    double elecrel[MAXSPECIES][MAXELECSTATE];   // inverse electronic collision numbers
```
stanmoore1, line 43: `double elecrel[MAXSPECIES][MAXELECSTATE]` is pretty big and will use up significant stack memory (12*16*8 bytes) and will bloat the Species struct when no electronic modes are defined. It may be better to define a new struct that is only allocated when electronic modes are defined. I'm open to Steve's opinion on this though.

src/surf_react_adsorb.h
```diff
@@ -192,6 +192,8 @@ class SurfReactAdsorb : public SurfReact {
     double v[3];            // particle velocity
     double erot;            // rotational energy
     double evib;            // vibrational energy
+    double eelec;
```
stanmoore1, line 195: New variables don't seem to be used?

stanmoore1 (commented):
@zseckert I started porting this to Kokkos but had a few comments after looking over the code. Thanks

src/create_particles.cpp
zseckert: This is almost certainly leftover from when the electronic energy was a new member of the particle class rather than using the custom data attribute. `fix_elecmode` has an update_custom method where we should set it, but apparently never propagated the `temp_elec` through. Is changing the signature on `fix::update_custom` to include `temp_elec` too invasive?

src/particle.h
zseckert: Yes, having a vector of nspecies inside each species struct is intended. Essentially each state and each collision partner species can have different relaxation numbers/behavior, hence the array. However, I rarely expect that array to be dense and unique (i.e. there will be many columns and/or rows which are filled with the same value). With your next comment we can almost certainly reorganize all this data to reduce the memory overhead.

src/particle.h
zseckert: That seems reasonable to me. Are you thinking essentially leave only `nelecstate` plus a pointer to the new struct in the `Species` struct, then we need only allocate that struct for the species where `nelecstate > 0`? Also, by delaying the allocation of that struct until slightly later (after we parse the electronic mode input file) we could allocate the vectors to only be as long as needed rather than guessing a maximum size. The parsing fits most naturally in `particle.cpp`, but would it be preferable to move the struct definition and the related methods to `fix_elecmode`? A similar method could be used to reduce the size of `elecrel`.

src/surf_react_adsorb.h
zseckert: Another leftover from before the custom data transition. A question I have though, does the call to `particle->add_particle` in line 1230 ever trigger the `update_custom`? It is easy to default to 0 energy in `update_custom` but we may have cruft left over if particles are being deleted/created quickly (this may be answered by our previous emails about vibrational energy).

src/create_particles.cpp
stanmoore1: >Is changing the signature on fix::update_custom to include temp_elec too invasive?

No I think it is reasonable, since it matches the other temps

stanmoore1:
@zseckert it seems like `elecstate` in `fix elecmode`, i.e. `int *elecstate = particle->eivec[particle->ewhich[elecstateindex]];` isn't required anywhere, is it needed?

src/particle.h
zseckert: @stanmoore1 I am working on this and most of the edits are straight forward. Essentially I made another struct that contains all the relevant information and only allocate it for the species that have electronic data defined. One difficulty I am running into is that during parsing in `particle.cpp`, I don't think I can trust `nspecies`. Specifically for that relaxation array where I want to store values for each species in the simulation and the following vector of bools. Would it be preferable to initialize them at some length and error if we need more space (i.e. using `MAXSPECIES`) or add logic to `grow_species` to also touch these vectors/arrays?

zseckert: Thanks for pointing that out. Since I am now allocating the memory after parsing the ElecFile it should be easy to remove those constants.

stanmoore1: @zseckert sounds good, after you push your changes I will port them to Kokkos

zseckert: @stanmoore1 I just pushed an update that dynamically allocates only the memory needed for the electronic mode. The MAXSPECIES and MAXELECSTATE constants are still floating around in the ElecFile struct that is used during parsing, which I will hopefully get to tomorrow, but all the memory that hasn't been released by the start of the simulation is hopefully in its final form. Let me know your thoughts. I still need to do a thorough search for dynamic array in stack memory.

zseckert: @stanmoore1 I have fixed a couple more dynamic arrays on the stack as well as modified the electronic state parsing to remove the constant sized arrays/vectors. The code for this is quite a bit messier with all the smalloc's and sfree's, but it's simple enough that I have confidence in it, and I have tested it all on my suite of examples that show the same results. I have now done everything that is on my list, unless I have overlooked something. Let me know what you need from me, but I think this is ready for you again.

src/collide_vss.cpp
```diff
@@ -550,17 +566,149 @@ void CollideVSS::EEXCHANGE_NonReactingEDisposal(Particle::OnePart *ip,
...
+  if (particle->species[ispec1].elecdat->species_rel[ispec2] != NULL) {
+    return particle->species[ispec1].elecdat->species_rel[ispec2][ielec];
+  } else {
+    return particle->species[ispec1].elecdat->default_rel[ielec];
+  }
+}
+
+/* ---------------------------------------------------------------------- */
+
+double CollideVSS::calc_elec_coll_temp(Particle::OnePart *p, double E_Dispose, double omega)
```
stanmoore1, line 632: @zseckert this function `CollideVSS::calc_elec_coll_temp` doesn't seem to be used, can it be removed?

src/particle.cpp
```diff
@@ -1089,6 +1221,70 @@ double Particle::evib(int isp, double temp_thermal, RanKnuth *erandom)
...
+
+    double ran = erandom->uniform();
+    ielec = 0;
+    while (ran > cumulative_probabilities[ielec])
+      ++ielec;
+  }
+  return ielec;
+}
+
+double Particle::eelec(int isp, double temp_elec, RanKnuth *erandom)
```
stanmoore1, line 1249: This function seems unused as well.

src/collide_vss.cpp
zseckert: Removed. It was originally intended for people who would want to implement temperature dependent relaxation models for electronic states, but the electronic temperature is very non-intuitive, so making them think hard about this before undertaking such a task is probably a good thing.

src/particle.cpp
```diff
@@ -933,7 +951,121 @@ void Particle::add_species(int narg, char **arg)
...
+      for (j = 0; j < nfile; j++)
+        if (strcmp(names[i],fileelec[j].id) == 0) break;
+      if (j == nfile) {
+        species[ii].elecdat = NULL;
+        continue;
+      }
+
+      int nmode = fileelec[j].nmode;
+      maxelecstate = MAX(maxelecstate, nmode);
+      species[ii].elecdat = new ElectronicData();
```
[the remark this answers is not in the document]
zseckert: Added logic to free it in the `Particle::~Particle`.

zseckert:
@stanmoore1 I think I have resolved all your more recent comments. The regression tests are passing here and I am going to run my longer tests locally with both the default and kokkos (CPU only) versions, but beyond that I'm not sure what else is needed.

stanmoore1:
@zseckert I'm still debugging Kokkos runtime issues, but if you can verify everything is working correctly with CPU (non-Kokkos) that would be great. If you have it would be good to run this once through Valgrind and check for memory issues or leaks

stanmoore1:
@zseckert I ran this through valgrind and fixed a invalid read, it is all clean now (and no leaks). I also fixed some other issues, and Kokkos is working on CPU, still need to test GPU.

zseckert:
@stanmoore1 My extended tests caught another error that was occurring at the ends of runs. I think I've fixed that and am rerunning my tests now. They take an hour or two, so I'll probably check them first thing in the morning and let you know if we are all good on that front.

stanmoore1:
GPU seems to be giving correct answer too, though performance may need to be optimized.

stanmoore1:
@zseckert can you add a doc page for `compute_telec_grid`? I need to port that to Kokkos as well.

aborner1:
I do not agree with this: https://github.com/sparta/sparta/blob/1b21a3fa3418ebbbde1df7beab520d38b5688507/src/react_tce.cpp#L85C57-L85C62
and same in react_tce_qk.cpp.

The electronic energy should not be used in the computation of the total energy in the TCE model (see Fig. 7.2 in Kyle Higdon's PhD dissertation).

mgallis:
So, when calculating ionization rates with TCE are you suggesting excluding the electronic energy?
It sounds counter intuitive.

Michael

From: Arnaud Borner ***@***.***>
Date: Thursday, March 14, 2024 at 4:32 PM
To: sparta/sparta ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [EXTERNAL] Re: [sparta/sparta] Implemented electronic excitation (PR #446)

I do not agree with this: https://github.com/sparta/sparta/blob/1b21a3fa3418ebbbde1df7beab520d38b5688507/src/react_tce.cpp#L85C57-L85C62
and same in react_tce_qk.cpp.

The electronic energy should not be used in the computation of the total energy in the TCE model (see Fig. 7.2 in Kyle Higdon's PhD dissertation).

aborner1:
Michael - I agree it is counterintuitive. In the general case, Kyle proved that using the electronic energy overpredicts the Arrhenius rate.
Screenshot 2024-03-14 at 3 59 51 PM
Furthermore, he says (page 175):

> "This approach cannot be extended to electronic excitation modeling since the observed errors are much greater for the much more widely spaced quantum electronic levels. The effects of including the electronically excited state energy in the TCE model can be seen for the O2 dissociation reaction in Fig. 7.2. For this reaction, the rate is overpredicted by an even larger margin than the case with rotational and vibrational energy contributing to the TCE mode. On a large scale, it was found that the ionization rates were severely overpredicted when the electronic energy was included in the TCE model. It is likely that this problem resulted in an overprediction of ionization in the second nominal simulation and sensitivity analysis (Chapters 4 and 5)."

zseckert:
@aborner1 That's a really interesting result. However, physically, as Michael said, that doesn't make intuitive sense. Electronically excited states take significantly less energy to ionize than the ground state and have different dissociation energies. I need to reread the TCE/QK sections in Bird's books again in more detail but I'm fairly certain that the derivation of reaction probability neglected electronic energy. It makes sense that it would overpredict the Arrhenius rate in that case as it is essentially just "extra" energy that TCE/QK wasn't designed to handle. Perhaps we need to revisit the derivation and generalize it?

@stanmoore1 FYI, my longer tests all passed.

mgallis:
If the calibration of the TCE parameters has included the electronic energy I cannot see why it would not reproduce the correct rates. In any case, we will confirm we are reproducing the correct equilibrium rates.

Michael

From: zseckert ***@***.***>
Date: Friday, March 15, 2024 at 7:55 AM
To: sparta/sparta ***@***.***>
Cc: Gallis, Michael A. ***@***.***>, Comment ***@***.***>
Subject: [EXTERNAL] Re: [sparta/sparta] Implemented electronic excitation (PR #446)

@aborner1<https://github.com/aborner1> That's a really interesting result. However, physically, as Michael said, that doesn't make intuitive sense. Electronically excited states take significantly less energy to ionize than the ground state and have different dissociation energies. I need to reread the TCE/QK sections in Bird's books again in more detail but I'm fairly certain that the derivation of reaction probability neglected electronic energy. It makes sense that it would overpredict the Arrhenius rate in that case as it is essentially just "extra" energy that TCE/QK wasn't designed to handle. Perhaps we need to revisit the derivation and generalize it?

@stanmoore1<https://github.com/stanmoore1> FYI, my longer tests all passed.

stanmoore1:
@sjplimp this is ready for your review

aborner1:
> @aborner1 That's a really interesting result. However, physically, as Michael said, that doesn't make intuitive sense. Electronically excited states take significantly less energy to ionize than the ground state and have different dissociation energies. I need to reread the TCE/QK sections in Bird's books again in more detail but I'm fairly certain that the derivation of reaction probability neglected electronic energy. It makes sense that it would overpredict the Arrhenius rate in that case as it is essentially just "extra" energy that TCE/QK wasn't designed to handle. Perhaps we need to revisit the derivation and generalize it?

I agree, and would be very happy to work on this together if you'd like.

stanmoore1:
@zseckert is `compute telec/grid` supposed to work with `species` output? The example seems to be give `NaN` if I try that.

src/compute_telec_grid.cpp
```diff
@@ -0,0 +1,345 @@
...
+    error->all(FLERR,"Illegal compute telec/grid command");
+
+  ngroup = particle->mixture[imix]->ngroup;
+  mixspecies = particle->mixture[imix]->nspecies;
+  nspecies = particle->nspecies;
+
+  ntally = 2*mixspecies;
+
+  per_grid_flag = 1;
+  size_per_grid_cols = ngroup;
```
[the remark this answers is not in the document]
zseckert: Yes, I agree with that.

zseckert:
> @zseckert is `compute telec/grid` supposed to work with `species` output? The example seems to give `NaN` if I try that.

It should work with `species`. Theoretically the calculation should also work with `all` but I don't think the calculation is physically meaningful. I'll take a look.

Also, I am going to push a commit that fixes some issues with relaxation after a chemical reaction. Essentially you must relax upon collisions to keep the electronic data in sync with other particle data.

zseckert:
@stanmoore1 I just added a second `compute telec/grid` to that example and I am getting identical results for both versions that appear correct. What did you do to produce `NaN`?

stanmoore1:
I will need to port the changes in [9b3b8e4](https://github.com/sparta/sparta/pull/446/commits/9b3b8e4fb08430ed82aa41866abf630673b89da4) to KOKKOS package.

zseckert:
After I pushed yesterday I realized I can pretty easily combine the two versions. I'll refactor later today as well.

zseckert:
@stanmoore1 I just pushed both the refactor of my commit from yesterday (should be much simpler to port) and a fix for the `species` flag. Thankfully that was very straightforward to fix, just a missing `else`.

stanmoore1:
@zseckert good catch, I think that was my mistake, thanks.

stanmoore1:
@zseckert does more work need to be done to address the concern by @aborner1 in this comment? https://github.com/sparta/sparta/pull/446#issuecomment-1998619451

zseckert:
> @zseckert does more work need to be done to address the concern by @aborner1 in this comment? [#446 (comment)](https://github.com/sparta/sparta/pull/446#issuecomment-1998619451)

@stanmoore1 I have some simulations running currently that should give us definitive evidence one way or the other, however from Michael's description, I don't think anything is needed. Essentially, because we input the raw Arrhenius rates into SPARTA rather than the TCE parameters (which SPARTA calculates internally), SPARTA should automatically adapt the coefficients to this new capability and not need further changes.

I'll check to see if my sims are done and try to put together a plot to show this.

stanmoore1:
@zseckert can you add an another example with reactions? I ported the changes for reactions to Kokkos but need to test. Thanks

stanmoore1:
@zseckert thanks, I'm seeing a regression with the Kokkos version + reactions, need to debug.

stanmoore1:
Kokkos test now is correct.

stanmoore1:
@zseckert I think this is ready to merge once you get the data back from the longer runs, and we get a final review from @sjplimp.

zseckert:
> @zseckert I think this is ready to merge once you get the data back from the longer runs, and we get a final review from @sjplimp.

@stanmoore1 I ran everything without the `partial_energy no` flag for reactions, so I am redoing the sims now. They should hopefully run tomorrow and I'll try to get plots up by the end of the week.

zseckert:
Here are the plots. Note solid lines include electronic energy, dashed lines do not. This is a 0D simulation starting with translational, rotational, and vibrational modes at 25,000 K in dry air, the electronic mode is originally not excited at all (100 K)

You can see from this first plot that the simulations including the electronic mode (solid lines) that the temperatures do not match because the energy is being split between an additional energy mode. These not matching is expected.
temps_25kK_short

But the chemistry, is very similar, but not perfectly the same:
specs_25kK_short
Most of the species agree very well but a few do not, specifically NO, N2+, and N+. This isn't statistical noise as I see the same trends on the same species across all my cases from 5,000 K up to 50,000 K.  I won't be able to get to it today, but I'll dig more into those specific reactions/species soon to see why the chemistry is diverging.

zseckert:
@stanmoore1 Sorry I got pulled away from this for a bit on other tasks. I found what was causing the discrepancy. In the TCE algorithm the reaction rate depends on the energy and the effective degrees of freedom. I had missed implementing the calculation of the electronic degrees of freedom in the algorithm for my previous results. I just pushed a commit that implements this. I also moved some electronic energy calculations from `compute_telec_grid` into `particle.cpp` since they are more generally useful.

Because the DoF between electronic and the other internal energy modes are different, I don't think we should expect the results to be exactly the same when partial_energy = false. I am rerunning the above cases and will post results when I have them, but I am expecting similar discrepancies as those shown above.

zseckert:
@stanmoore1 I know I've been quiet here for a month or two but I've been picking away at this as I got time. I've merged master into this branch and also implemented one more change that I think finalizes this feature (assuming people agree with the decision I've made, described below).

Essentially in the TCE model for chemistry, we required an effective dof for the electronic mode, exactly like we have for the discrete vibrational mode. In the discrete vib mode, we assume that the vib energy in the reacting particle is the average energy over all particles and calculate a vibrational temperature that is consistent with that mean energy. I tried doing the same thing for the electronic mode, however I ran into some issues. Specifically, because the electronic mode has a finite number of states, particles that were in the higher energy states (top 1-2) would have such a high mean energy that the temperature calculation and effective dof calculation wouldn't make sense. For example, calculating the dof for N2 as you step up through a simplified electronic structure with 8-ish states would give reasonable effective dof for the lower states (between 0 and 3, although they non-intuitively decreased as the state increased) but the 7th state would give a small negative dof and the last state would give something absurd like -300. These negative dof would then cause the chemistry to go haywire if any particles ever reached these states.

I did some digging in the literature to see if anyone had a solution to this problem. I found a couple references but they included cell averaging. I also tried brushing up on my quantum mech to see if an argument could be made to set the dof based on spin or degeneracy, but didn't fully convince myself. I then decided to simply add another input parameter to the electronic input file so that the effective dof for the TCE model could simply be set by the user. This has a couple advantages:

1. It is more performant. We are just reading a number from memory instead of doing a newton/bisection solve of the distribution function
2. It allows easy experimentation with the various methods I attempted above (as well as any others people can think of)
3. It should allow relatively easy modification of per-state chemistry rates. I know there is sparse data for some excited states and some reactions, so the values could be tuned to reproduce that data.

However this obviously leaves more for the user to figure out in setting up problems. I've done some simulations with values between 0 and 2, and the differences are on the same order as what I have plotted above. I'll push a commit shortly that updates the documentation for the new parameter in the electronic state file, then, I think it is back to you to migrate these changes to kokkos.

stanmoore1:
@zseckert thanks for letting me know, sounds good

stanmoore1:
@zseckert I know it has been a long time, but with AI's help I now have this code updated and nearly ready to merge. Can you please run your tests once more to make sure nothing broke?

zseckert:
> @zseckert I know it has been a long time, but with AI's help I now have this code updated and nearly ready to merge. Can you please run your tests once more to make sure nothing broke?

I've pulled this down and am running things now. Some of my tests are needing updating for various reasons, but I'll try to knock this out quickly.

zseckert:
I have my smaller tests updated and everything looks reasonably correct (all the electronic modes are equilibrating as expected). I've currently got my larger simulations running that produced the chemistry plots above, however they will take some time. I'll try to get an update here either tomorrow or early next week once they complete.

stanmoore1:
Sounds good, thanks @zseckert!

zseckert:
@stanmoore1 My runs finally finished (I should have picked a faster machine to run them on). Temperatures and species densities are within statistical noise of each other for the entire simulation duration so resolving the merge conflicts didn't affect behavior in any way I can see.

sjplimp (approved):

PR #447, opened 2024-01-10
Added functionality to convert explicit surfaces to implicit surfaces…

andrewykhong:
## Purpose

This pull request adds functionality to SPARTA to read in an explicit surface and convert it into an implicit surface. The primary purpose is to effectively allow ablation of explicit surfaces. 

## Author(s)

Andrew Young Kyun Hong from Sandia National Laboratories. Helpful discussions about implementation with Steve Plimpton and Stan Moore.

## Backward Compatibility

No backward compatibility issues known.

## Implementation Notes

These changes primarily generate an array which derives the appropriate corner point values from a user-defined explicit surface. These corner point values are passed to a "fix_ablate" which then generates the implicit surfaces. All functions are contained within create_isurf command and the corresponding files. 

A new command was added called create_isurf. The documentation for this new command details the format and functionality.

One minor change was made to the function "check_uniform_group" in grid.cpp to account for split cells.

## Post Submission Checklist

None at the time


andrewykhong:
@stanmoore1 I have replaced all tabs with spaces. I have also added the necessary document. I added some pictures to the /doc/JPG for the document page. These images reflect the most recent version which has been tested for more complicated surfaces in both 2d and 3d.

stanmoore1:
@andrewykhong great I will take a look. Is this ready to merge from your point of view?

efrainhr:
@andrewykhong came across this request looking for this capability! (thanks @stanmoore1 for pointing it out)

Tried the examples under the exp_imp directory. They all run when executing in serial; however all (but the box example) crash with watertight errors when running in parallel. Here's an example error message (for egg)

ERROR: Watertight check failed with 348 unmatched edges (/home/sparta/src/surf.cpp:1576)

Other errors were: duplicate edges (cone; line 1368), and unmatched point (cyl; line 1137). Running on a cray (i.e., aprun -n 2 sparta_exe < input_file).

andrewykhong:
@stanmoore1 @efrainhr  I am aware there is a bug when this is run in parallel. I'm working on this right now. I hope to get a version which is ready to merge soon.  The code should always work in serial. 

If you would like to use the converted surface in parallel, there is a workaround. In serial, you may read in an explicit surface and convert  it to an implicit surface using "create_isurf". Then you can "write_isurf" to output the implicit surface. That implicit surface can then be run in parallel. This workflow is not ideal and should not be necessary after I push some changes over the next week.

andrewykhong:
@stanmoore1 There is a bug in the code when run in parallel. This version is not ready to merge. I'll comment when it is ready.

efrainhr:
> @efrainhr I am aware there is a bug when this is run in parallel. I'm working on this right now. I hope to get a version which is ready to merge soon. The code should always work in serial.
> 
> If you would like to use the converted surface in parallel, there is a workaround. In serial, you may read in an explicit surface and convert it to an implicit surface using "create_isurf". Then you can "write_isurf" to output the implicit surface. That implicit surface can then be run in parallel. This workflow is not ideal and should not be necessary after I push some changes over the next week.

great, thanks for the tip! I'll try it out

andrewykhong:
@stanmoore1 I believe this version of the explicit to implicit surface conversion is working. I've tested the cases in example/exp_imp/debug, example/exp_imp/egg, and example/exp_imp/cone for multiple procs. Have not encountered the watertight check fails or gridcut error. The documentation has also been updated

stanmoore1 (changes requested):
Hi @andrewykhong this is looking good. I had a few change requests:

- Can you rename `sparta/examples/exp_imp` to something a little more descriptive, either "explicit2implicit" or "explicit_implicit"?
- Can you remove the "debug" folder from the examples, or rename it to something like "simple"?

Thanks

src/grid.cpp
```diff
@@ -2441,7 +2443,12 @@ int Grid::check_uniform_group(int igroup, int *nxyz,
   bigint bcount = count;
   MPI_Allreduce(&bcount,&allbcount,1,MPI_SPARTA_BIGINT,MPI_SUM,world);
 
-  if ((bigint) nxyz[0] * nxyz[1]*nxyz[2] != allbcount)
+  bigint allscount;
+  bigint scount = splitcount;
+  MPI_Allreduce(&scount,&allscount,1,MPI_SPARTA_BIGINT,MPI_SUM,world);
+
+  // some inconsistency here if surface file is large (comment out for now)
```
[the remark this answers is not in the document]
andrewykhong: This is an old comment and no longer applicable. The check_uniform_group function should always be consistent

src/create_isurf.cpp
```diff
@@ -0,0 +1,1560 @@
...
+  MPI_Barrier(world);
+
+  // find remaining corner values based on neighbors
+  if (me == 0)
+    if (screen) fprintf(screen,"Cleanup corners ...\n");
+
+  int full;
+  if(dim==2) full = find_side_2d();
+  else full = find_side_3d();
+  if(!full) error->all(FLERR,"could not fill");
```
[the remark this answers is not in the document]
andrewykhong: I changed the error message to be more specific. Is the new one ok?

src/create_isurf.cpp
```diff
@@ -0,0 +1,1560 @@
...
+  MPI_Barrier(world);
+
+  if (me == 0)
+    if (screen) fprintf(screen,"Syncing corners ...\n");
+
+  sync(CVAL);
+  MPI_Barrier(world);
+
+  // free up memory since these are no longer needed
+  //memory->destroy(ivalues);
```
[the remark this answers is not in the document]
andrewykhong: Old comment which I have removed

andrewykhong:
@stanmoore1 I have made the requested changed. I also removed some inputs in the examples/explicit2implicit folder, but there are still three there so should be enough for users to become familiar with the new function

stanmoore1:
Sounds good, thanks @andrewykhong

src/create_isurf.cpp
```diff
@@ -0,0 +1,1636 @@
...
+    error->all(FLERR,"Create_isurf thresh must be bounded as (0,255)");
+
+  // mode to determine corner values
+
+  if (strcmp(arg[3],"inout") == 0) aveFlag = 0;
+  else if (strcmp(arg[3],"ave") == 0) aveFlag = 1;
+  else error->all(FLERR,"Unknown surface corner mode called");
+
+  //if (aveFlag && comm->nprocs > 1)
+  //  error->all(FLERR,"Create_isurf averaging not possible in parallel");
```
stanmoore1, line 144: Remove or enable commented out code

src/create_isurf.cpp
```diff
@@ -0,0 +1,1636 @@
...
+  MPI_Barrier(world);
+
+  // remove old explicit surfaces
+
+  remove_old();
+
+  surf->implicit = 1;
+  surf->exist = 1;
+
+  tvalues = NULL; // TODO: Add per-surface type
```
stanmoore1, line 166: TODO: should this done before merging?

doc/create_isurf.txt
```diff
@@ -0,0 +1,162 @@
...
+part of exactly 2 triangles.  Exceptions to these rules allow for
+triangle edges (in 3d) that lie entirely in a global face of the
+simulation box, or for line points (in 2d) that are on a global edge of
+the simulation box.  This can be the case after clipping, which allows
+for use of watertight surface object (e.g. a sphere) that is only
+partially inside the simulation box, but which when clipped to the box
+becomes non-watertight, e.g. half of a sphere. Generated implicit surfaces will
+always be watertight.
+
+While not currently implemented, implicit surface elements can have a
```
stanmoore1, line 54: Not sure what this means. Can you please clarify or remove.

stanmoore1 (changes requested):
@andrewykhong I reviewed again and have a few more questions/change requests.

doc/create_isurf.txt
andrewykhong: I removed the statement in the most recent commit

src/create_isurf.cpp
andrewykhong: Right now, all surface elements are assigned to the same surface type. Assigning per-grid-cell surface types was not a capability I was planning to implement right now. Maybe in the future if it's requested?

sjplimp:
@andrewykhong I looked over everything except the 4 source files.  I will work on those next.  Here are some suggestions on the rest of the files:

(1) If you edit a doc/*.txt file you need to run the txt2html converter tool to also update the *.html files.  I did that for what I just committed.

(2) I reworked the create_isurf doc page.  Please read it (HTMLversion is there) and see if you agree with my changes.  Also I added a few NOTE comments, which you should respond to or make corresponding edits.  Also, I made the included images smaller, on the HTML page you can click on them for the large version.

(3) I think you should add an examples/explicit_to_implicit dir and the various input/output files you added to the other dirs: cone/circle/sphere in the new dir instead.

(4) Not clear why coarse_sphere.surf was added to the tools dir?  If it is needed by the example input scripts, it should be in that dir. If it is meant to illustrate something about implicit surfs, it should probably be in the data dir, similar to surf.apollo (for example).

stanmoore1:
>(3) I think you should add an examples/explicit_to_implicit dir with the various input/output files you added to the other dirs cone/circle/sphere in the new dir instead.

@sjplimp just to clarify, do you mean we should collapse the existing `examples/explicit2implicit/cone,circle,sphere` directories into a single `examples/explicit_to_implicit` folder?

sjplimp:
Sorry on suggestion #3 - I misread the directory structure in the PR, thinking you had added a create_isurf example to each of the top level cone, circle, sphere dirs, not as sub-dirs of explict2implicit.

However, since the 3 sets of inputs/outputs are either the same (air.vss) or disjoint, having them in the same top-level explicit2implicit dir probably makes the most sense.  There are other example dirs with several input scripts on the same theme.

andrewykhong:
@sjplimp @stanmoore1 
(1+2) I modified the doc files and corrected typos and converted them to html.
(3) I removed the subdirectories in examples/explicit2implicit. They are all now together. Updated the species and vss files so that they all run. I'm seeing some log files. Should these be deleted?
(4) I removed the surf file from tools. This was here during testing and should have been removed.

stanmoore1:
>I'm seeing some log files. Should these be deleted?

No these are the "correct" answer for the `circle` example used for regression testing. I tried to add `sphere` and `cone` as well but they appeared to be running out of memory and crashing on CI runner machine, so they are currently disabled.

andrewykhong:
Also updated the summary for this pull request (I don't know what it's called other than the big block of text at the top).

sjplimp:
@andrewykhong Here is my 2nd batch of suggestions

(1) I made some more edits to your new version of doc/create_isurf.txt.  See the several ANDREW comments.  Also added some simple diagrams for how I think implicit surfs can be different than explicit to illustrate the accuracy bullets.  See if you think they are correct?

(2) Simplified Grid::check_uniform_group() a bit.  You should verify it still works as you expect.

(3) I think the new examples dir needs a simple README text file.  Since it appears there are 5 different input scripts, not all with exp2imp in the name?  What are the 3 cone variants?  Also, each of the input scrips should have a pair of log files (1,4 procs).  If this is not possible for memory or CPU time reasons, then the examples are too big.  Ideally they should each run in under 1 minute, even on 1 proc.

(4) I made some mainly cosmetic edits to create_isurf.cpp.  Look at all my ANDREW comments and Qs.   After addressing all the comments, you should verify the final version works as you expect on all your tests cases, including the examples dir.

(5) Have you run all your tests with valgrind, both serial and parallel?  To make sure there are no errors or mem leaks.

(6) Added your name to the top of the create_isurf.cpp file - good job!  This is a complex and powerful new feature.

andrewykhong:
@sjplimp 
(1) I slightly modified the diagrams so they are more similar sized and clearly discriminate the implicit and explicit surface. Also added example for bullet point 1

(2)  I ran the inputs from the examples/explicit2implicit and did not see the error being thrown because of the split cells. In other words, the function is still working as intended

(3) I removed the cone examples because I couldn't reduce their size to be sufficiently small. Reduced the size of the sphere input. There are now three inputs: 2D circle, 3D sphere, and 3D sphere with ablation. I added a README file to explain each input. 

(4) Updated create_isurf.cpp. Removed unncessary comments and updated descriptions to functions

(5) I ran all cases on valgrind both in serial and parallel. There are no errors from either the create_isurf.cpp or grid->check_uniform_group function. 

(6) Thank you!

andrewykhong:
@stanmoore1 
The log files in the examples/explicit2implicit directory may need to be updated

sjplimp:
@andrewykhong thanks for cleaning up the diagrams for the doc page and adding a new one, as well as addressing my other comments.  

My latest commit simply tweaks the paragraph describing the diagrams.  And adds a sentence as to why the new example diagram does not have a blue line which overlays the red one, since it is non-intuitive.  Is that sentence correct?  I guess for the corner diagram, something like that would have also been the case, except I happened to draw it with the intersection points at the same fractional distance along the 2 grid cell edges?  I will review-to-accept, and you can still change that paragraph if you like.

sjplimp (approved):

andrewykhong:
@sjplimp I corrected two typos in the documentation but otherwise, the sentence is correct. 
@stanmoore1 I saw the tests are now failing. I'll take a look at these Monday.

stanmoore1 (approved):

PR #451, opened 2024-01-15
Add Kokkos support for FFTs

stanmoore1:
## Purpose

Add Kokkos support for FFTs in SPARTA, including `compute fft/grid`. This also removed an unnecessary source of non-determinism in the Kokkos `irregular` communication routine.

## Author(s)

James Almgren-Bell (UT Austin), Sam Mish (UC Davis), and Stan Moore (SNL)

## Backward Compatibility

Removes support for older FFTW2 library; FFTW3 is still supported.


stanmoore1:

PR #461, opened 2024-03-12
Stochastic weighted particle method

andrewykhong:
## Purpose

Implements the stochastic weighted particle method (SWPM) to SPARTA.  SWPM is a generalization of the DSMC method where particle weights are no longer constant but stochastically change over the simulation time. SWPM is useful when there is a large variance in the local number of simulators.

## Author(s)

Andrew Young Kyun Hong.

## Backward Compatibility

N/A

## Implementation Notes

Particle weights are stored in the "weight" attribute in the particle class. This attribute is currently only used in axi-symmetric simulations. Currently, SWPM should not be used with axi-symmetric simulations (error throws have been added), and there does not seem to be only conflict. The particle restart has an additional attribute to store the particle weight in the restart files.

The splitting and reduction algorithms for SWPM are included in the new collide_swpm.cpp file. In collide_vss.cpp, the test_collision(..) and perform_collision(..) have minor changes to account for the variable particle weights. The computes have also similar changes. 

If SWPM is not explicitly enabled with a collide_modify command, the implementation has no affect on SPARTA. If SWPM is enabled, only the particle collisions are affected. The particles move and collide with surfaces as DSMC particles.

Implementation was tested for various cases, some of which are compiled in the following paper:
"An improved stochastic weighted particle method for boundary driven flows"
https://doi.org/10.1016/j.jcp.2024.113353

I have also included a python script to generate particle velocities according to the BKW distribution. This was mainly used in testing the swpm implementation and is a fairly common test for variable weight DSMC methods. This python file may be omitted.

I have also included an additional compute called "compute_vmom_grid.cpp" which computes the raw velocity moments. This was used to measure from the BKW relaxation problem (referred above), but this new compute could be omitted as well. I have added it the documentation.

Documentation has also been updated to reflect all changes.

## Post Submission Checklist


andrewykhong:
I had to include math_eigen.cpp, math_eigen.h, math_eigen_impl.h from LAMMPS for the stochastic weighted particle method. For SPARTA to compile, I needed cpp version 11 so  I had to add two compiler flags "-O3" and "-std=c++11".  Should I add in the documentation that these compiler flags are needed or add them to each MAKE file? Some guidance here would be appreciated.

doc/collide_modify_swpm.txt
```diff
@@ -0,0 +1,130 @@
...
+collide_modify keyword values ...  :pre
+
+one or more keyword/value pairs may be listed :ulb,l
+keywords = {swpm} or {reduce}
+  {swpm} values = choice Nsplit gamma
+    choice = {yes} or {no} to allow particle weights to stochastistically change
+    Nsplit = maximum number of particles in a cell to split particles
+    gamma = weight transfer functino which is a value > 0
+  {reduce} values = reduction_choice Nreduce group_choice Ngroup
+    reduction_choice = {ENERGY} or {HEAT} or {STRESS} = how to reduce particles
```
stanmoore1, line 22: We typically don't have input deck options as all caps, e.g. ENERGY, HEAT, STRESS, can you please make these lowercase instead?

doc/collide_modify.txt
```diff
@@ -135,6 +135,8 @@ vibrational mode, then use of the {discrete} option also requires the
 "fix vibmode"_fix_vibmode.html command be used to allocate storage for
 the per-particle mode values.
 
+There is also an option to use the stochastic weighted particle method (refer to "collide_modify_swpm"_collide_modify_swpm.html)
```
stanmoore1, line 138: It seems awkward to have an orphaned option from the same command in a different file, but I'm open to discussing this.

src/collide_vss.cpp
```diff
@@ -765,6 +772,105 @@ void CollideVSS::EEXCHANGE_ReactingEDisposal(Particle::OnePart *ip,
 
 /* ---------------------------------------------------------------------- */
 
+int CollideVSS::split(Particle::OnePart *&ip,
```
stanmoore1, line 775: Is `CollideVSS::split` unique to VSS, or could it be moved to `Collide` instead?

src/math_eigen.cpp
```diff
@@ -0,0 +1,78 @@
+/* ----------------------------------------------------------------------
```
stanmoore1, line 1: We need to decide if we should put the SPARTA copyright on these files with a note that they are copied from LAMMPS, or if we should keep the LAMMPS copyright, which seems a little weird in SPARTA.

src/particle.h
```diff
@@ -78,6 +79,7 @@ class Particle : protected Pointers {
     int flag;               // used for migration status
     double dtremain;        // portion of move timestep remaining
     double weight;          // particle or cell weight, if weighting enabled
+    double g;               // particle weight (= FNUM if DSMC used)
```
stanmoore1, line 82: We typically discourage adding new members to the `OnePart` struct, prefer using the `custom` feature instead. However I'm open to @sjplimp's opinion on this.

Also using `g` for a variable is not descriptive and self obvious (at least to me), is there a different longer name that would be more descriptive?

stanmoore1 (changes requested):
@andrewykhong thanks for all the work on this, it is looking like good progress. I had some comments and questions:

- Can you please rename the `in.debug` and `in.heat_debug` input scripts in the examples to be more descriptive?
- Regarding C++11 support, we have discussed making that a requirement for SPARTA and I think now is a good time. That may be better done in a separate PR though, it would need CMake updated too in addition to all the Makefiles.

andrewykhong:
@stanmoore1 For the C++11 support, you mentioned that we should have a separate pull request. For the Makefiles, would this more or less need to have the c++11 flag added to each one?

stanmoore1:
> @stanmoore1 For the C++11 support, you mentioned that we should have a separate pull request. For the Makefiles, would this more or less need to have the c++11 flag added to each one?

Correct. We'd also need to update CMake, documentation, and ideally have a check in the code so that the user knows their compiler needs to be updated, rather than it just failing to compile.

andrewykhong:
@stanmoore1 I finished some testing this morning and I think this is good (or at least until someone finds a bug). I updated the doc to. I also removed the octree grouping . It's not needed and was more to check results during testing

andrewykhong:
@stanmoore1 there's a bug I'm trying to fix. I'll recomment when this is ready. Do not merge this yet

andrewykhong:
@stanmoore1 I finished moving all the particle weights within the "weight" attribute in the particle class as we have discussed. I removed the fix_swpm since it's seems superfluous. All checks could be moved into collide_modify and there are no custom particle arrays needed now for SWPM.

I have added a new compute (compute_vmom_grid) and a python script to generate a certain distribution (bkw_init.py). These two files aren't necessary for the implementation and can be omitted.

This PR is ready to be reviewed.

andrewykhong (commented):
All requested changes made

src/particle.h
```diff
@@ -173,6 +175,7 @@ class Particle : protected Pointers {
   void unpack_restart(char *, int &, int, int);
 
   int find_custom(char *);
+  double *stochastic_weights();
```
andrewykhong, line 177: @stanmoore1 this was based on a recommendation by Claude. I think it's harmless but not 100% sure

andrewykhong:
@stanmoore1 I finally found some time to clean up this PR. You may need to bless the SWPM example. I tested with examples/swpm. I used Claude to help push and from what I can tell, didn't do anything offensive. I'd like to finally get this merged; have some time these next few weeks

PR #463, opened 2024-03-19
Modified small corner value removal in fix_ablate

andrewykhong:
## Purpose

Modified how small corner point values are adjusted. Currently, corner point values are adjusted by a constant "epsilon" which may have a marginal effect depending on the choice of threshold value. Instead, the corner point values should be adjusted such that the vertices of the surface elements do not lie too close to each corner. Otherwise, "skinny" or degenerate triangles may be created. To achieve this, the "isosurface stuffing" is implemented where corner point values are adjusted to zero out any corner points which would lead to these "skinny" triangles. Only a few lines of code were added. This is also an optional argument that a user can define.

Also some changes made to create_isurf to take care of the warning during compile time

## Author(s)

Andrew Young Kyun Hong

## Backward Compatibility

N/A

## Implementation Notes

A few lines of code were added to fix_ablate.  The default value for "alpha_low" which defines the smallest distance between the intersection and the corner point is 0.02. This value is relative to the cell length. 

## Post Submission Checklist

No new examples were added.


stanmoore1 (changes requested):
A few comments.

doc/fix_ablate.txt
```diff
@@ -225,3 +242,8 @@ This fix can only be used in simulations that define implicit surfaces.
 "read isurf"_read_isurf.html
 
 [Default:] none
```
stanmoore1, line 244: Please list the default value for the new keyword, for example see https://github.com/sparta/sparta/blob/master/doc/fix_ave_time.txt#L298-L302

src/create_isurf.cpp
```diff
@@ -171,7 +171,7 @@ void CreateISurf::command(int narg, char **arg)
   
   tvalues = NULL;
   int pushflag = 0;
-  char *sgroupID = arg[0];
+  char *sgroupID = 0;
```
stanmoore1, line 174: Please set this to `NULL` instead of `0`.

stanmoore1 (changes requested):
A few comments.

stanmoore1:
@andrewykhong I'm also seeing a few compiler warnings from `create_isurf`, could you fix these in this PR too?

```
../create_isurf.cpp:266:13: warning: data argument not used by format string [-Wformat-extra-args]
            allofull);
            ^
../create_isurf.cpp:397:46: warning: expression result unused [-Wunused-value]
            else if (svalues[icell][i] == 2) 0; // do nothing
                                             ^
../create_isurf.cpp:410:46: warning: expression result unused [-Wunused-value]
            else if (svalues[icell][j] == 2) 0; // do nothing
                                             ^
../create_isurf.cpp:544:46: warning: expression result unused [-Wunused-value]
            else if (svalues[icell][i] == 2) 0; // do nothing
                                             ^
../create_isurf.cpp:556:46: warning: expression result unused [-Wunused-value]
            else if (svalues[icell][j] == 2) 0; // do nothing
```

stanmoore1:
@andrewykhong great thanks, I'll rebless the gold logs so the tests pass.

doc/fix_ablate.txt
andrewykhong: Updated documentation to indicate default value

src/create_isurf.cpp
andrewykhong: Made this change

stanmoore1:
@andrewykhong this PR is causing `ablation` examples to fail with numerical regressions, in addition to the `explicit2implicit` examples. Is this the expected behavior?

andrewykhong:
@stanmoore1 That should not be happening. In the current version, the corner point values which are "epsilon" above the threshold are reduce to the threshold hold value minus epsilon. The only change I made was the value of epsilon itself (it was too small before). I'll take a look. 

Are the explicit2implicit tests still failing? From the log files for the tests, I only see the ablation cases failing.

stanmoore1:
I reblessed the `explicit2implicit` examples so they are not failing.

stanmoore1:
@andrewykhong thanks I will rebless the tests.

stanmoore1 (approved):

andrewykhong:
Fixed a conflict between versions

sjplimp:
@andrewykhong The explanation of the new keyword was confusing to me on the fix ablate doc page.  So I rewrote the explanation to be more clear.  However, I am not sure my explanation is correct.  Can you check it ?

Also, I changed the name of the keyword from surfbuffer to mindist, since "buffer" in the manual typically refers to chunks of memory.

Also, I left a NOTE comment on the doc page, asking about the old style of using a pre-defined EPSILON.

Finally, in fix_ablate.cpp I don't understand the formulas for these lines of code and how they ensure the min distance of a tri vertex from the grid corner point, which was my understanding when I reworked the doc page explanation:
    cmin = (thresh - 0.0 * mindist) / (1.0 - mindist);
    cmax = (thresh - 255.0 * mindist) / (1.0 - mindist);
Can you explain them to me?

andrewykhong:
@sjplimp I read through the fix ablate doc and it looks good. The old epsilon adjust based on (thresh-epsilon) is still in effect. The default is actually mindist = 0.0. In this case, it reverts to the old epsilon_adjust

The equations are the interpolation equations in marching cubes/square rearranged. For convenience, the interpolation equation is:
value = lo + (hi - lo) * (thresh - v0) / (v1-v0)
The lower corner point has a value of v0; the upper corner point has a value of v1. Since mindist is relative to the cell length, I set lo = 0 and hi = 1. Mindist is equal to the value or the distance of the vertex from the lower corner point. The equation can be rearranged to solve for v0 (which is the corner point being adjusted):
v0 = (thresh - v1 * mindist) / (1.0 - mindist)
There are two cases:
1) [ v0 < thresh ]. I assume v1 is 255.0. Then v0 must be less than cmax for the vertex to be at least mindist away from the corner.
cmax = (thresh - 255.0 * mindist) / (1.0 - mindist)

2) [ v0 > thresh]. I assume v1 is 0.0. Then v0 must be greater than cmin for the vertex to be at least mindist away from the corner.
cmin = (thresh - 0.0 * mindist) / (1.0 - mindist)

The cmin and cmax are meant to be understood as the minimum value an inside corner point can be AND the maximum value an outside corner point can be. I understand this is confusing so I switched the terminology so that (cmin < cmax)

I hope that makes sense

andrewykhong:
Made a few changes to make the code easier to read. I believe it's ready for review

sjplimp:
Thanks @ayhong for the further details.  I think I understand it now.  I changed the doc page again to more carefully explain the algorithm and also to explain the connection between the default epsilon and isosurface stuffing algorithms.

I also changed the src code to use mindist and commented what it is doing more fully.  I also renamed cmin and cmax in the header and source file to hopefully be more self-explanatory:  corner_inside_min and corner_outside_max.

If you agree with these changes and I didn't break aything, I will approve the PR for merging.

andrewykhong:
@sjplimp I agree with the changes. I like the doc page explanation. I made a few very minor changes in the doc.

sjplimp (approved):

PR #491, opened 2024-06-21
Add custom per-surf options to fix emit/surf

sjplimp:
## Purpose

Enable fix emit/surf to emit particles from surfaces using custom per-surf values for density, temperature, and vstream (or speed).
This allows customization of the emission properties of different surface elements.  The per-surf values can be defined by the read_surf or custom commands.

## Author(s)

Steve

## Backward Compatibility

New feature, hopefully does not break any existing features.

## Implementation Notes

## Post Submission Checklist

### DIFF
--- a/doc/fix_emit_surf.html
+++ b/doc/fix_emit_surf.html
@@ -37,7 +37,7 @@
     Psub = pressure setting at inflow boundary (pressure units)
     Tsub = temperature setting at inflow boundary, can be NULL (temperature units)
   <I>custom</I> values = attribute s_name
-    attribute = <I>nrho</I> or <I>vstream</I> or <I>speed</I> or <I>temp</I>
+    attribute = <I>nrho</I> or <I>vstream</I> or <I>speed</I> or <I>temp</I> or <I>fractions</I>
     s_name = custom per-surf vector or array with name 
 </PRE>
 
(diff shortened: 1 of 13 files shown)


sjplimp:
This needs testing from Michael - also a few loose ends to decide on.

zseckert:
@sjplimp Michael asked me to test this and he set up a case with a cube surface file I had generated. He may have sent you the file, but in short it had emission from the -x surface and lesser emission from the -y surface. I got what appeared to be correct results running in serial, but using more than 1 core output seemingly from random tris on the surf.

I did some digging, but was unable to 100% confirm what was happening. Here is my understanding so far:

Each processor only stores 1/nproc of the triangles in a given surface when the file is read
If the surface is not distributed, later this data is shared between all the processes (I believe during the Grid::surf2grid call). I think every processor should end up with all the tris/custom data with the index in edvec equal to tris[i].id = i +1. I am not completely aware of the difference between edvec and edvec_local, so that may play a role.
During Grid::surf2grid, in surf_comm.cpp line 270-271 inside Surf::rendezvous_redistribute_surfs you see:

 270     if (distributed) m = (id-1) / nprocs;
 271     else m = (id - nsurf_old - 1) - me*surfperproc

       The second line is the one called, and it essentially uses id (which is the tri id) minus 1 as the index. This is what places the triangles into the triangle vector in the appropriate location
        4) However, in the Surf::rendezvou_redistribute_custom method around line 364, you only see:
         j = (id-1) /  nprocs
        i.e. the distributed version of the code.

Before surf2grid, for example when running with 2 processors, rank 0 would get the custom data for the odd id'ed triangles while rank 1 would get the even id'ed triangles during parsing of the file, but then the above communication would then garble the data further. I tried changing this to mimic the earlier code but was getting double free errors and segfaults then, so there is obviously something I don't understand about the memory layout/manipulation that happens.

If you have pointers or further things you want me to dig into, it would be greatly appreciated.

sjplimp:
@zseckert - pretty sure it's an issue with the owned vs local custom per-surf data - I saw it yesterday when I was working on adding a species fraction option - hopefully I will get it fixed today

zseckert:
@sjplimp My qualitative checks seem to produce the right answers. I'll work to generate a more complex input and test it out.

stanmoore1:
I looked over this PR and I don't see anything that needs to be ported to Kokkos.

stanmoore1 (approved):
@sjplimp said this is ready to merge.

PR #495, opened 2024-07-10
Better implicit convert

andrewykhong:
## Purpose

This pull request introduces 1) improvements to the explicit to implicit surface conversion and 2) a new multi-point decrement for DSMC ablation.  

1) The marching cubes/squares algorithm has been updated to interpolate vertices from inner values. Inner values are an array of values stored at each corner. Previously, only one value is stored at each corner. The advantage is the intersection between the explicit surface elements and the cell edges can be perfectly copied during the surface conversion. I have also added a "voxel" option which is similar to "ave" but determines the corner point by averaging the cell volume around the 2x2(x2) stencil around each inside corner point

2) The multipoint decrement fundamentally changes how SPARTA does ablation. Previously, only the minimum corner point is reduced each ablation cycle. This causes anomalous surface recession at the cell-level scale. There is also a bias associated with the order in which the corner points are checked. The multi-point decrement, as the name suggests, updates multiple corner points each ablation cycle and does not suffer from the ordering bias. Additional routines are included which allow both the multi-point decrement and the inner indices to be used at the same time.

These changes only affect's SPARTA's ability to simulate ablation. If no fix_ablate or create_isurf command is specified, SPARTA's previous functionality does not changed. 

Additional things added:

A new ablation source is also added called uniform. The "uniform" option applies a uniform cell decrement to all cells which contain part of the surface and gas. 

I have also included an additional python tool to convert SPARTA surfaces back to STL files which can be read by software such as SOLIDWORKS and MeshLAB.

I also fixed a typo in one of the doc pages and included a link in the section_commands doc so users can find the create_isurf command

## Author(s)

Andrew Young Kyun Hong (SNL)

## Backward Compatibility

None 

## Implementation Notes

Changes were made to fix_ablate, the marching square/cubes files, and create_isurf. The inner indices were test by converting a 3D sphere, cone, and capsule; the implicit surfaces were then used in a short DSMC simulation with no ablation. The multipoint decrement is tested by simulating a 2D etch and a 3D fiber. The first case is ablated with the uniform ablation and the second case is ablated using gas-surface reactions whose probability is a constant. These results along with more test cases can be found in the publication "Towards physically realistic ablation in direct simulation Monte Carlo simulations" (in progress)

## Post Submission Checklist

(Will add DOI later)


src/create_isurf.cpp
```diff
@@ -808,22 +947,26 @@ void CreateISurf::comm_neigh_corners(int which)
...
+    memory->create(inghost,maxghost,ncorner,ninner,"createisurf:inghost");
     memory->create(cghost,maxghost,ncorner,"createisurf:cghost");
     memory->create(sghost,maxghost,ncorner,"createisurf:sghost");
-    memory->create(ighost,maxghost,ncorner,nadj,"createisurf:ighost");
+    memory->create(ighost,maxghost,ncorner,ninner,"createisurf:ighost");
   }
 
   // unpack received data into val_ghost = ghost cell corner points
 
+  // NOTE: need to check if hashfilled
```
stanmoore1, line 968: @andrewykhong is this comment still relevant?

stanmoore1:
@andrewykhong overall this looks good to me, I had one question about a comment in the code. Is it supposed to break the ablation regression tests though?

src/create_isurf.cpp
andrewykhong: Nope. I removed the comment in a commit just now

andrewykhong:
@stanmoore1 I added one more input to the examples/ablation called in in,ablation.multi.inner.3d. Could this be why they are failing?

stanmoore1:
>I added one more input to the examples/ablation called in in,ablation.multi.inner.3d. Could this be why they are failing?

No that shouldn't cause failures. It looks like the numerics of the existing tests changed at little:

```
==== comparing log.mpi_1.ablation.2d with log.7Feb24.mpi_1.ablation.2d ====
Step     error  0.0 wrt norm   500.0
Np       error  3.0 wrt norm 158002.0
Natt              error  0.0
Ncoll             error  0.0
Nscoll   error  2.0 wrt norm   128.0
Nscheck  error 396.0 wrt norm  8
f_bal    error  0.0 wrt norm     1.0
f_bal[2]  error  0.0 wrt norm     1.0
f_ablate  error  4.0 wrt norm 717570.0
!!! test ablation.2d FAILED
elapsed time = 2.209153890609741 s for test ablation.2d
```

stanmoore1:
I can rebless the tests if you think the output is correct, we just need to double check there were no unintended changes that could point to a bug.

andrewykhong:
@stanmoore1 FYI, I'm on this right now. I'll be pushing a few more commits tomorrow and possibly the next day. I'll add another comment to follow up soon

andrewykhong:
@stanmoore1 I reverted some sections of the marching_cubes/squares and all the ablation tests are passing. The explicit2implicit sphere example is good but is failing. I think it's because of the following: 

1) There is a warning line in the old log files because the number of particles created is less than expected (this is normal because the surfaces are slightly distorted following the surface conversion) . The updated create_isurf has smaller thresholds and the number of particles does not mismatch . In short, there is one extra line in the old log files that is trivial

2) Similar to previous point, the thresholds were slightly changed which is causing the surfaces to be slightly adjusted. This causes a slight change in the surface volume which leads to a difference in the number of surface collisions.  But general behavior is overall unchanged.

In short, I think the explicit2implicit example is actually ok and should be reblessed.

stanmoore1:
@andrewykhong is this PR ready for review? Also can you please fix the few merge conflicts with latest master?

andrewykhong:
> @andrewykhong is this PR ready for review? Also can you please fix the few merge conflicts with latest master?

I'll fix the merge conflicts / issues.

andrewykhong:
@stanmoore1 This is ready for a review. 

The explicit2implicit example needs to be reblessed. I had to make the domain a little larger for the "voxel" option to not fail. The "voxel" option causes the surface to slightly expand which pushes the edges of the sphere outside the domain. There was also a small bug in the sync in create_isurf. I double checked that if I did not fix the bug and used the old example input, the tests do pass (commit [1a0b70a] is an example of this)

andrewykhong:
@stanmoore1 Hi Stan, I'm looking into a potential bug. I'll followup if it turns out not to be a bug or when I fix it

andrewykhong:
I fixed a bug in the voxel-based surface conversion which changed the results for the exp2imp.sphere example. Could you rebless that log? @stanmoore1

stanmoore1:
> I fixed a bug in the voxel-based surface conversion which changed the results for the exp2imp.sphere example. Could you rebless that log? @stanmoore1

Yes will do.

src/marching_cubes.cpp
```diff
@@ -502,14 +503,6 @@ void MarchingCubes::cleanup()
   MyPage<surfint> *csurfs = grid->csurfs;
   int nglocal = grid->nlocal;
 
-  // DEBUG
```
stanmoore1, line 505: These debugging statements can be useful, please restore.

stanmoore1 (commented):
@andrewykhong here are a few comments, overall this looks good

doc/create_isurf.txt
```diff
@@ -15,9 +15,13 @@ create_isurf group-ID ablateID thresh mode ... :pre
 group-ID = grid ID for which grid cells to perform calculation on :ulb,l
 ablateID = ID of a "fix ablate"_fix_ablate.html command :l
 thresh = threshold for corner values used to generate implicit surfaces, value > 0.0 and < 255.0 :l
-mode = {inout} or {ave} :l
+mode = {inout} or {ave} or {multi} :l
   {inout} = mark corner points as either inside or outside surf
-  {ave} = smooth values by averaging expected corner point values based on intersections between the explicit surfaces and cell edges :pre
+  {ave} = smooth values by averaging expected corner point values based on intersections between the explicit surfaces and cell edges 
+  {multi} = mark corners using multivalues :pre
```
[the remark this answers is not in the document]
andrewykhong: We used the work multivalue in the paper. 

What about the following change?
"mark corners using multiple values (i.e. multivalues)"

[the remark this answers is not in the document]
stanmoore1: Yes that sounds good

src/marching_cubes.cpp
andrewykhong: Added back these lines.

doc/create_isurf.txt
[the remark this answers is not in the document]
andrewykhong: Updated the doc. Also added another paragraph to explain new voxel option

stanmoore1:
@sjplimp Andrew said this is ready for review, no more changes coming

src/decrement_lookup_table.h
```diff
@@ -0,0 +1,1114 @@
+/**
+ * @file    LookUpTable.h
+ * @author  Thomas Lewiner <thomas.lewiner@polytechnique.org>
+ * @author  Math Dept, PUC-Rio
```
stanmoore1, line 4: @andrewykhong does this file have a license?

examples/ablation/log.12Nov24.mpi_1.ablation.multi.inner.3d
```diff
@@ -0,0 +1,267 @@
...
+stats_style		step cpu np nattempt ncoll nscoll nscheck
+
+run 100
+Memory usage per proc in Mbytes:
+  particles (ave,min,max) = 0 0 0
+  grid      (ave,min,max) = 1.51379 1.51379 1.51379
+  surf      (ave,min,max) = 0.0117035 0.0117035 0.0117035
+  total     (ave,min,max) = 1.86455 1.86455 1.86455
+Step CPU Np Natt Ncoll Nscoll Nscheck 
+       0            0        0        0        0        0        0
```
stanmoore1, line 114: @andrewykhong this example has all zeros for stats output. Is there some variable we could use to check the ablation is working? Can use `compute reduce ave` to output a single value to the stats.

src/decrement_lookup_table.h
andrewykhong: I'm not sure. This was implemented before this PR. I believe Arnaud added this to SPARTA? @aborner1

examples/ablation/log.12Nov24.mpi_1.ablation.multi.inner.3d
andrewykhong: I'll change the stats style to show the total volume (there's a compute reduce with the volumes) and comment once more after I'm done

src/decrement_lookup_table.h
aborner1: Well we've had the original lookup_table.h file in there for years. I had emailed Thomas Lewiner in 2019, and here is a snippet of his response:
```
Regarding the Marching Cubes license, it is 100% free (even no GPL: I did not know which one to choose at the time...).
```

examples/ablation/log.12Nov24.mpi_1.ablation.multi.inner.3d
stanmoore1: @andrewykhong sounds good, thanks

src/decrement_lookup_table.h
stanmoore1: @aborner1 thanks, I added this to the file.

examples/ablation/log.12Nov24.mpi_1.ablation.multi.inner.3d
andrewykhong: updated the stats_style so it outputs the total gas volume over time @stanmoore1

stanmoore1 (approved):

stanmoore1:
@andrewykhong one question, I see that there is `doc/JPG/multi_single_ablation_small.png` but no larger version (i.e. non-small), is that intended?

examples/ablation/log.12Nov24.mpi_1.ablation.multi.inner.3d
stanmoore1: Thanks!

andrewykhong:
> @andrewykhong one question, I see that there is `doc/JPG/multi_single_ablation_small.png` but no larger version (i.e. non-small), is that intended?

Yes. There should just be one picture

andrewykhong:
@stanmoore1 Pushed new changes. Only the in.exp2imp.circle test cases are failing, but this is expected since the degenerate triangle handling was updated. Grid corner points are not changed but the vertices are shifted in marching cubes/squares. No longer is there any small mass removal and the vertices should now always be distant from the grid corner points

stanmoore1:
@sjplimp can you please review this PR

tools/surf2stl.py
```diff
@@ -0,0 +1,106 @@
+#!/bin/bin/env python
```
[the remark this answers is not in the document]
sjplimp: I added this to the commit I made for "small doc page changes"

sjplimp:
@andrewykhong @stanmoore1 I think this is ready to merge from my perspective - I made a few minor changes to the doc pages, but otherwise everything else looks good!

@andrewykhong Do you want to add a citation to your new PoF paper on the create_isurf and fix_ablate doc pages ?

andrewykhong:
Added citation to paper

PR #502, opened 2024-07-29
Surf emit momentum

zseckert:
## Purpose

These changes include the momentum and energy contributions of `fix emit/surf` in the results of `compute surf`. 

## Author(s)

Zak Echo, Sandia National Laboratories

## Backward Compatibility

All previous inputs should work with the new functionality, however simulations that use both `fix emit/surf` and `compute surf` on the same surface will produce different results

## Implementation Notes
These changes call `surf_tally` on all relevant computes from `fix emit/surf`. Since it is a particle creation event, `iorig` is left NULL while `ip` is the new particle, allowing all the previous logic to work for accumulating particle properties. Most of the changes in `compute surf` are to support the NULL value of `iorig`, as it previously assumed that it always existed.

I checked the problem, in parallel, with a cube that emitted different flow values from two different sides, ensuring that the forces and mass flux was correct.

## Post Submission Checklist


stanmoore1:
@zseckert @sjplimp what is the status of this PR?

mgallis:
Ready to merge.

Michael
________________________________
From: Stan Moore ***@***.***>
Sent: Monday, August 26, 2024 3:01:23 PM
To: sparta/sparta ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [EXTERNAL] Re: [sparta/sparta] Surf emit momentum (PR #502)


@zseckert<https://github.com/zseckert> @sjplimp<https://github.com/sjplimp> what is the status of this PR?

sjplimp:
@stanmoore1 I'd like to tweak a couple things

stanmoore1:
> @stanmoore1 I'd like to tweak a couple things

@sjplimp OK let me know when you are done.

sjplimp:
@zseckert I made some small changes to the logic in compute_surf.cpp.  Also commented out the dump file write in the examples script (shouldn't be produced unless the user wants it).  If you have some test cases to run, it would be good to verify no answers have changed.  @stanmoore1 Otherwise ready to merge

zseckert:
@stanmoore1 @sjplimp I ran my test and the results were consistent with the previous results. This looks good from my perspective.

stanmoore1 (approved):

PR #524, opened 2024-10-08
Add fix halt, copied and adapted from LAMMPS

### LINKED ISSUE #519: Missing "run N time" command functionality
## Summary

Hi,

The current SPARTA user documentation (https://sparta.github.io/doc/run.html) suggests that a simulation can be set to run for a specified amount of physical time, rather than a fixed number of timesteps, by use of the "time" keyword. However, this functionality does not appear in the present code version.

Searching around the GitHub, I have found a few prior commits that appear to have added this functionality (e.g. https://github.com/sparta/sparta/commit/41d36015b621fc0ffb4714d95fe780c871d9ad3a), but these modifications do not appear in the change logs for the relevant files in the master branch.

I was wondering if there was a reason for this functionality not currently existing in the master branch. This would be useful when submitting SPARTA simulations to an HPC cluster with restrictive run time limits (i.e. 8 hours) that necessitate frequent restarts/resubmissions.

Thanks!

## Type of Issue

Question/Suggestion

4Sep2024

From the SPARTA manual for the "run" command: 

"Examples:
run 10000
run 1000000 upto
run 100 start 0 stop 1000
run 1000 pre no post yes
run 100000 start 0 stop 1000000 every 1000 "print 'Temp = $t'"
run 100000 every 1000 NULL
run 5.7 time
run 5.7 time every 100 NULL 

Description:
Run or continue a simulation for a specified number of timesteps. If the time keyword is used, run for a specified amount of physical time."

Attempting to use the "time" keyword currently returns the following error message: "ERROR: Illegal run command (../run.cpp:102)"

Replace the run command from any example file (e.g. in.collide) with the command "run 5.7 time" from the above documentation.

N/A

### LINKED ISSUE #522: Add run option to automatically identify steady state
## Summary

Many simulations should be run "to steady state/equilibrium," so it would be convenient if a run command could be specified to run until some value had reached a plateau and then for some physical time after than.

## Type of Issue

'Suggestion for an Enhancement'

Along the lines of https://github.com/sparta/sparta/issues/519 but with a different angle, I often find myself running a simulation a few times for different durations and manually examining output quantities to identify how long it takes to reach a steady state. This is especially tedious/impractical when launching a grid of simulations for a parameter scan, where each one may require a different number of timesteps to equilibrate (well, stabilize - the steady state may be far from physical equilibrium). 

I would like to suggest a new keyword (perhaps `until`?) for the `run` command that allows the user to specify the following:

- A quantity to monitor, like total number of particles, temperature, average speed of particles tallied on a surface, or some other compute
- A gradient (fractional change per time) threshold, like 1% per 10 seconds. When this is reached, a new stats keyword (perhaps `steady`) would change from 0 to 1.
- A sampling window, like 100 timesteps or 0.1 s of physical time. Values of the monitored quantity would be stored in a ring buffer along with their timestamps, and the gradient would be computed at each timestep from the values in the ring buffer.
  - Optionally, there might be cases where it is unnecessary to store the monitored quantity at every timestep. Instead, one could use an `every` argument to sample more sparsely.
- A post-run time (to accumulate statistics) after the stability condition is reached. 
  - This should be able to trigger the start of new outputs (like dump files or images). I'm not sure how to handle that yet. As a temporary solution, the user could filter all the normal outputs by looking for `steady` = 1.
  - In its simplest form, the post-run duration would be a number of steps or amount of physical time. 
  - Even more useful would be a counting/statistical threshold like reaching a certain number of particles tallied on a target surface. 
  - A particularly sophisticated version of the stopping condition would be to track a quantity (or its variance), like the mean speed of particles in a volume. Statistics for this second quantity would only start to accumulate after satisfying the steady-state condition. The most flexible way to specify this would be as a formula.
- Finally, it may be useful to have an over-riding condition, which might be a number of steps, a physical time, or - as in https://github.com/sparta/sparta/issues/519 - a wall time.

Once the specified condition is reached, the run would continue for the statistics-gathering period and then terminate.

The steady-state tracking quantity (either the present value of the gradient, or the present value divided by the target value) should also be a stats option, so the user can observe progress in the terminal. Using the `(present/target)` approach would allow this single value to both track progress towards steady state _and_ serve as the Boolean indicator of whether steady state has been reached (is `(present/target)` smaller than 1?).

An example might be

`run until Np gradient 0.01 per 1 seconds window 10 seconds every 20 steps then_until "c_1[3] > 1000" except_max 1000000 steps`

This would run the simulation until the fractional gradient of the total number of particles per 1 s was less than 1%, based on a ring buffer 20 seconds long containing samples every 20 timesteps. Then, having reached steady state, the run would continue until the compute c_1[3] was larger than 1000, and then end. If the steady-state condition was never reached, the run would terminate after 1000000 steps.

I've included units (`seconds` and `steps`) so it's clear if a time-like quantity is one or the other; there should be a third option for wall time as well.

I suppose some of this behavior may already be present in the LAMMPS `fix halt` that was discussed in #519, but unfortunately I'm not a LAMMPS user so I don't know.

There might be communication overhead in parallel runs depending on what quantity is being tracked, since some quantities require collecting information from all sub-processes. I don't have a clear idea of how significant a bottleneck this might be, but I hope it's small since the communication is one way and only a single value is handed up.

I suggested a similar tool for MolFlow+ a few years back, which was implemented and is now widely used. I know Sparta doesn't have a GUI (and probably shouldn't), but I think the suggestion above will provide the same basic function - with the added benefit of automatic termination.
https://molflow-forum.web.cern.ch/t/automatic-convergence-testing-and-prediction/398
https://gitlab.cern.ch/molflow_synrad/molflow/-/commit/a75fdb89d4ebd5f79ad85fd6e590eeedebc14088

stanmoore1:
## Purpose

Add `fix halt`, copied and adapted from LAMMPS. This fix allows one to stop the simulation after a user-specified walltime limit, or based on the output from a variable.

## Author(s)

Code was adapted to SPARTA by Stan Moore (SNL), original LAMMPS code was written by Axel Kohlmeyer (Temple U)

## Backward Compatibility

Yes

Closes #519
Closes #522

### DIFF
--- a/doc/fix_halt.html
+++ b/doc/fix_halt.html
@@ -79,6 +79,14 @@ stop the run after an hour of walltime:
 <PRE>variable cpu equal cpu
 fix 10 halt 10 v_cpu > 3600.0 
 </PRE>
+<P>The commands above apply only to the time spent in the current run
+command. If multiple run commands are used in the same input script,
+one can also stop the run after a predetermined amount of <I>total</I>
+walltime:
+</P>
+<PRE>variable wall equal wall
+fix 10 halt 10 v_wall > 3600.0 
+</PRE>
 <P>Similarly one can stop the run after a predetermined amount of
 simulation time, which is useful when using a <A HREF = "fix_dt_reset.html">variable
 timestep</A>:
@@ -134,7 +142,9 @@ files</A>.
 
 <P><B>Restrictions:</B> none
 </P>
-<P><B>Related commands:</B> none
+<P><B>Related commands:</B>
+</P>
+<P><A HREF = "run.html">run</A>
 </P>
 <P><B>Default:</B>
 </P>
--- a/doc/fix_halt.html
+++ b/doc/fix_halt.html
@@ -71,13 +71,20 @@ properties, grid properties, surface properties, global values
 calculated by a <A HREF = "compute.html">compute</A> or <A HREF = "fix.html">fix</A>, or
 references to other <A HREF = "varible.html">variables</A>. Thus they are a very
 general means of computing some attribute of the current system.  For
-example, the following two versions of a fix halt command will do the
-same thing:
+example, the following two versions of a fix halt command will both
+stop the run after an hour of walltime:
 </P>
 <PRE>fix 10 halt 10 tlimit > 3600.0 
 </PRE>
 <PRE>variable cpu equal cpu
 fix 10 halt 10 v_cpu > 3600.0 
+</PRE>
+<P>Similarly one can stop the run after a predetermined amount of
+simulation time, which is useful when using a <A HREF = "fix_dt_reset.html">variable
+timestep</A>:
+</P>
+<PRE>variable time equal time
+fix 10 halt 10 v_time > 1.0e-3 
 </PRE>
 <P>The choice of operators listed above are the usual comparison
 operators. The XOR operation (exclusive or) is also included as "|^".
--- a/doc/fix_halt.html
+++ b/doc/fix_halt.html
@@ -1,0 +1,136 @@
+<HTML>
+<CENTER><A HREF = "https://sparta.github.io">SPARTA WWW Site</A> - <A HREF = "Manual.html">SPARTA Documentation</A> - <A HREF = "Section_commands.html#comm">SPARTA Commands</A> 
+</CENTER>
+
+
+
+
+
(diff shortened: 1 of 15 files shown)


stanmoore1:
Still need to adapt the documentation

PR #526, opened 2024-10-14
Update surf2paraview.py

### LINKED ISSUE #464: write_surf flag + surf2paraview
## Summary

If the "write_surf" command is used with the "type" flag, then surf2paraview fails.

## Type of Issue

Bug Report

## SPARTA Version

Current master

surf2paraview.py should be general enough to detect whether there is an additional integer (the surface element type) and discard it.

Surf2paraview returns an error due to one too many integers.

```
Error reading SPARTA surf input file
Triangles or Lines section contains bad line:  1 1 1 2 3
```

Read any input file in an example (let's say, sphere), then write it, with the "type" flag turned on. Then try and postprocess the surface with surf2paraview.

This is a very old bug. The recent addition of the "type" flag in "write_surf" actually provided a way to avoid the error, but, for years, I have a carried a modified version of surf2paraview for surface files written out by SPARTA.

timteichmann:
## Purpose

This fixes #464. 

While #499 was intended to fix this issue, it introduced a new bug. The bug is that the type id was read as one of the point ids. This is because the type is written to the 2nd field.

The present pull request amends this by reading the point ids in reverse order from the end of the line.

Note: while this now supports type, it will fail catastrophically if write_surf is used with custom arg. There is no adhoc solution to support the general case because it would involve unknown number of fields (1 or 2 with type) before the point ids and unknown number of fields after. Solution would have to be to introduce an argument to surf2paraview if custom arg is used and provide the number of fields so that the script can infer the correct position of the point ids in the Lines list.

## Author(s)

Tim Teichmann

## Backward Compatibility

No (fixes bug in previous version)

## Implementation Notes

Trivial

## Post Submission Checklist

### DIFF
--- a/tools/paraview/surf2paraview.py
+++ b/tools/paraview/surf2paraview.py
@@ -97,7 +97,7 @@ def read_elements(sif, num_points, num_elements, ug, three_d_file):
 
   if ug.GetNumberOfCells():
     print("Error reading SPARTA surf input file")
-    print("Trangles or Lines section of file occurs more than once")
+    print("Triangles or Lines section of file occurs more than once")
     sys.exit(1)
 
   for line in sif:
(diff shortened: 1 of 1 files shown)


timteichmann:
@tjotaha please have a look and comment. Maybe you have a better fix or want to work on a more general solution.

tjotaha:
@timteichmann Thanks for fixing this error. My unit test for this case passes but I neglected to check the output surface in ParaView for correctness.

stanmoore1 (approved):
@timteichmann thanks for the bugfix!

PR #528, opened 2024-10-21
Work around roundoff issues in axisymmetric code

### LINKED ISSUE #527: "Collision cell volume is zero" error
## Summary

For some simulations, in particular 2D axisymmetric ones, it is possible to trigger a "Collision cell volume is zero" error.

## Type of Issue

Bug Report

## SPARTA Version

Latest master as of October 16th, 2024, 6:15pm EDT. [6c1199b](https://github.com/sparta/sparta/commit/6c1199b4bdd2af4fe49024eb9591615e63c7b5f9).
However, this bug has existed for a long time.

The code should elegantly skip the cells which volume is 0.0, or too close to machine precision 0.

The code returns an error triggered at line 484 in collide.cpp:

> ERROR on proc 5: Collision cell volume is zero (../collide.cpp:484)

Attached input script and surface file. air.species and air.vss from /data folder. Please run in parallel on 8 threads (should be reproducible due to "comm/sort yes" option). Error should be triggered at step 42319. Takes roughly 30 seconds on 8 threads. Output is also attached.

A known fix is to add the following line in collide.cpp at line 482:
`if (volume < 1.e-30) continue;`
However, that simply bypasses the problem.

I have verified that the volume of the cell that creates the problem is indeed exactly 0.0, and it contains 6 particles.
I suspect that particles collide with surface which belongs to cell that have a 0.0 volume (i.e., grid cells which are just touched by a surface).

[collision_cell_bug_input.txt](https://github.com/user-attachments/files/17403389/collision_cell_bug_input.txt)
[axi_surf.txt](https://github.com/user-attachments/files/17403351/axi_surf.txt)
[output.txt](https://github.com/user-attachments/files/17403392/output.txt)

stanmoore1:
## Purpose

Work around roundoff issues in axisymmetric code that cause particles to escape into surfaces.
Fixes #527

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

May change trajectories of previous bad cases

## Implementation Notes

Tested on reproducer input from #527

### DIFF
--- a/src/KOKKOS/geometry_kokkos.h
+++ b/src/KOKKOS/geometry_kokkos.h
@@ -171,6 +171,20 @@ bool axi_horizontal_line(double tdelta, double *x, double *v,
     nc = 1;
   }
 
+  // roundoff can cause code to miss collisions
+  //  occurs when t is epsilon greater than tdelta
+  //  set t = tdelta in this case
+
+  if (fabs(t1 - tdelta) < EPSSQ) {
+    if (tdelta > 0.0 && t1 > tdelta) {
+      t1 = tdelta;
+    }
+  } else if (fabs(t2 - tdelta) < EPSSQ) {
+    if (tdelta > 0.0 && t2 > tdelta) {
+      t2 = tdelta;
+    }
+  }
+
   // require first collision time >= 0.0 and <= tdelta
 
   if (t1 < 0.0 || t1 > tdelta) {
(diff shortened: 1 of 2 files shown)


stanmoore1:
Needs to be ported to Kokkos.
May be a better fix https://github.com/sparta/sparta/issues/150 for than the workaround in https://github.com/sparta/sparta/pull/221, I need to test.

stanmoore1:
I looked back at my emails, and it appears this PR would fix https://github.com/sparta/sparta/issues/150. Based on the debugging output, it has the same method of failure (`t` is epsilon greater than `tdelta`, leading to a missed collision in `axi_horizontal_line`). However I don't have a reproducer to know 100%.

PR #529, opened 2024-10-24
Enable variables in fix ave/histo to work with KOKKOS package

stanmoore1:
## Purpose

Enable variables in fix ave/histo to work with KOKKOS package, which previously was not supported.

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

Yes


stanmoore1:
This appears to be working correctly

PR #530, opened 2024-10-25
Renumber implicit surfaces in `write_surf` since ablation may remove surfs

### LINKED ISSUE #525: Implicit Surfaces written to a surface file using "write_surf" cannot be read back into SPARTA using "read_surf"
## Summary

Explicit surface geometry files generated from implicit surfaces via the "write_surf" command cannot be read back into SPARTA as-is using the "read_surf" command due to non-sequential surf_id values.

## Type of Issue

Bug Report

4Sep2024

From the manual for the write_surf command: "Write a surface file in text format describing the currently defined surface elements, whether they be explicit or implicit surfaces. See the [read_surf](https://sparta.github.io/doc/read_surf.html) and [read_isurf](https://sparta.github.io/doc/read_isurf.html) commands for a definition of surface elements and how they are defined and used by SPARTA. The surface file can be used for later input to a new simulation or for post-processing and visualization."

When the write_surf command is called, the code currently writes out the stored surf_id as the index for a surface triangle (or assumably the surface line segment in 2D) in the generated explicit surface file. However, for implicit surfaces, these surf_ids do not appear to be assigned sequentially, or with an initial value of 1, as expected by the "read_surf" command, resulting in an error. Manually changing the triangle IDs in the surface file to begin with a value of 1 and increase sequentially to n = # of total triangles causes the geometry to load correctly.

Place the modified in.ablate.3D input file in the "ablate" example directory and run it to generate a surface file. Running explicit_read.txt should then return the following error: "ERROR: Read_surf minimum surface ID is 1101 (../read_surf.cpp:1999)". Tested with both spa_serial and spa_mpi.

[explicit_read.txt](https://github.com/user-attachments/files/17311684/explicit_read.txt)
[in.ablate.3d.txt](https://github.com/user-attachments/files/17311685/in.ablate.3d.txt)

stanmoore1:
## Purpose

Renumber implicit surfaces in `write_surf` since ablation may remove surfs. Also document existing restriction that `write_surf` `custom` keyword cannot be used with implicit surfaces.

Fixes #525

## Author(s)

Stan Moore (SNL)

## Backward Compatibility
Yes

### DIFF
--- a/doc/write_surf.html
+++ b/doc/write_surf.html
@@ -134,7 +134,9 @@ next 3 processors and write it to a surface file.
 </P>
 <HR>
 
-<P><B>Restrictions:</B> none
+<P><B>Restrictions:</B>
+</P>
+<P>The <I>custom</I> keyword cannot be used with implicit surfaces.
 </P>
 <P><B>Related commands:</B>
 </P>
--- a/doc/write_surf.txt
+++ b/doc/write_surf.txt
@@ -127,7 +127,9 @@ next 3 processors and write it to a surface file.
 
 :line
 
-[Restrictions:] none
+[Restrictions:]
+
+The {custom} keyword cannot be used with implicit surfaces.
 
 [Related commands:]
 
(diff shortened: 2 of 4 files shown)


stanmoore1:
I tested the inputs from #525 and they still give the expected behavior.

PR #531, opened 2024-10-28
Two bugfixes

stanmoore1:
## Purpose

- Only renumber surface ids in `write_surf` for the file, but not in the actual data structures since this can cause other issues. (This was feedback from @sjplimp). This was changed recently in #530.
- Fix a bug in `dump image` that led to seg fault

## Author(s)

Stan Moore (SNL), and Steve Plimpton (retired SNL)

## Backward Compatibility

Yes

### DIFF
--- a/src/dump_image.cpp
+++ b/src/dump_image.cpp
@@ -1259,7 +1259,7 @@ void DumpImage::create_image()
 
     Compute *cx,*cy,*cz;
     Fix *fx,*fy,*fz;
-    int ppgflagx;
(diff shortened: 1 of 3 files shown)


stanmoore1:
I tested the inputs from https://github.com/sparta/sparta/issues/525 and they still give the expected behavior.

stanmoore1:
@sjplimp I think I implemented what we talked about previously here. Do you want to review, or should I go ahead and the bug fixes merge now?

PR #533, opened 2024-10-30
Fix rare issue with `Grid::point_outside_surfs_explicit` function

stanmoore1:
## Purpose

Fix rare issue with `Grid::point_outside_surfs_explicit` function. This issue caused particles to be created inside surfaces with `create_particles cut yes` and also gave false positives with `fix grid/check outside yes`, saying particles were inside surfaces when they were actually not.

When a surface was clipped to the cell bounds, if the intersection was tiny, the point "outside" the surfs could get pushed too far and actually end up inside nearby surfaces. The solution is to use the surface with the largest clipped area instead of the first surface, and push off the centroid of the entire polygon formed by the clipped surface instead of the center of the first triangle in the polygon. This should have no performance impact except when creating particles and when using `fix grid/check outside yes` which is already optional. `Fix ablate` uses implicit surfaces so it is not affected.

## Author(s)

Stan Moore (SNL), in discussion with Steve Plimpton (SNL retired)

## Backward Compatibility

Yes

## Implementation Notes

Only seen with the "spiky sphere" example (not in the public repo).

### DIFF
--- a/src/grid_surf.cpp
+++ b/src/grid_surf.cpp
@@ -1920,6 +1920,9 @@ int Grid::point_outside_surfs_explicit(int icell, double *x)
   double minsize = MIN(hi[0]-lo[0],hi[1]-lo[1]);
   double displace = EPSSURF * minsize;
 
+  double maxarea = 0.0;
+  int setflag = 0;
+
   if (dim == 2) {
     int npoint;
     double cpath[4];
@@ -1944,10 +1947,18 @@ int Grid::point_outside_surfs_explicit(int icell, double *x)
         if (edge == 4 and norm[1] > 0.0) continue;
       }
 
+      // use the surface with the largest clipped surface area
+      // for surfaces with a tiny interection, the point can get pushed
+      //  too far and end up inside nearby surfaces
+
+      double area = Geometry::poly_area(npoint,cpath);
+      if (area < maxarea) continue;
+      maxarea = area;
+
       x[0] = 0.5*(cpath[0]+cpath[2]) + displace*norm[0];
       x[1] = 0.5*(cpath[1]+cpath[3]) + displace*norm[1];
       x[2] = 0.0;
-      return 1;
+      setflag = 1;
     }
 
   } else {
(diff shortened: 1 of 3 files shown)


stanmoore1:
@sjplimp this is ready for review. I changed it to use the centroid of the entire polygon which should be more robust than using the center of first triangle like we did previously.

stanmoore1:
The spiky sphere example now runs without issues, including with `fix grid/check outside yes`.

src/grid_surf.cpp
```diff
@@ -1976,20 +1990,30 @@ int Grid::point_outside_surfs_explicit(int icell, double *x)
         if (face == 6 and norm[2] > 0.0) continue;
       }
 
-      double onethird = 1.0/3.0;
-      x[0] = onethird*(cpath[0]+cpath[3]+cpath[6]) + displace*norm[0];
-      x[1] = onethird*(cpath[1]+cpath[4]+cpath[7]) + displace*norm[1];
-      x[2] = onethird*(cpath[2]+cpath[5]+cpath[8]) + displace*norm[2];
-      return 1;
+      // use the surface with the largest clipped surface area
+      // for surfaces with a tiny intersection, the point can get pushed
```
sjplimp, line 1994: for surfaces with a tiny intersection, the point to push off from can be very close to a shared edge
  with another triangle - if the angle between the two tris is less than 90 degrees, the pushed-off
  point can end up "inside" the surface instead of "outside"
using the triangle with the largest clipped area and pushing off its centroid makes this
  issue very unlikely to occur

sjplimp (changes requested):
just tried to make the comment more clear - otherwise good to go on merge

src/grid_surf.cpp
```diff
@@ -1944,10 +1948,20 @@ int Grid::point_outside_surfs_explicit(int icell, double *x)
         if (edge == 4 and norm[1] > 0.0) continue;
       }
 
+      // use the surface with the largest clipped length
+      // for surfaces with a tiny intersection, the point can get pushed
```
sjplimp, line 1952: could add here a similar enhanced comment to the 3d triangle one

PR #534, opened 2024-10-31
Fix bug in `fix_ave_histo` and `fix_ave_histo_weight` with mixtures

stanmoore1:
## Purpose

Fix bug in `fix_ave_histo` and `fix_ave_histo_weight` leading to particles not being tallied correctly when using the mixtures with the `mix` keyword.

## Author(s)

Stan Moore (SNL), reported by Michael Gallis (SNL)

## Backward Compatibility

Yes

## Implementation Notes

Tested to fix an input from Michael


sjplimp (approved):
Good catch Stan

PR #540, opened 2024-11-15
use mean collision time from compute lambda/grid for variable timestepping

akstagg:
## Purpose

This variable time step improvement uses the more accurate mean collision time calculated in compute lambda/grid rather than calculating mean collision time internally (using the mean free path from compute lambda/grid) in compute dt/grid.

## Author(s)

Alan Stagg, SNL

## Backward Compatibility

This PR requires that the "tau" option (mean collision time) be used in the compute lambda/grid input command.

## Implementation Notes

The mean collision time from compute lambda/grid is used directly in compute dt/grid. Other changes there involve a more stringent evaluation of cell data for a cell timestep to be calculated (e.g., mean collision time, temperature, and speed must all be positive) and the epsilon comparison was removed.

## Post Submission Checklist

- [*] The feature or features in this pull request is complete
- [*] Suitable new documentation files and/or updates to the existing docs are included
- [*] One or more example input decks are included
- [*] The source code follows the SPARTA formatting guidelines


stanmoore1:
Thanks @akstagg! I will review.

stanmoore1 (approved):

PR #549, opened 2025-03-03
Fix int32 overflows in Irregular communication

stanmoore1:
## Purpose

Fix int32 overflows in Irregular communication for huge problems that lead to cryptic MPI errors. Allow the total send and receive buffers in irregular communication to be larger than 2 GB. However note that an *individual* send buffer in MPI still cannot be larger than 2 GB due to restrictions in MPI, but using `global mem/limit` can work around this in many cases.

## Author(s)

Stan Moore (SNL), issue reported by Matt Bettencourt (NVIDIA). Part of the overflow fix was inspired by LAMMPS `irregular.cpp`.

## Backward Compatibility

Yes


stanmoore1:
@bathmatt

stanmoore1:
I discussed this with @sjplimp, ready to merge

PR #550, opened 2025-03-04
New fix custom command along with new custom comand options

sjplimp:
## Purpose

This PR implements several new features:

(1) Enable custom command to read values from a file.  

(2) Add a new fix custom command with syntax similar to the custom command, but which allows for resetting custom attribute values from a variable or series of files, periodically as a simulation runs.

Both (1) and (2) work with the fix emit/surf command to allow tailored per-surf emission of particles, either from an object in the flow, or from a boundary specified by surface elements.  The fix custom command allows the emission to be time-dependent.

(3) Added a new action option called file/coarse to both the custom and fix custom commands.  This allows per-grid custom attributes to be defined by reading a file which lists a collection of coarse grid points, each with custom attribute values.  The coarse grid points could be output by continuum code (e.g. Navier Stokes) which computes a solution on a coarse grid.  SPARTA finds the closest coarse grid point to each SPARTA grid cell center point, and assigns the attribute values for the coarse grid to be the per-grid custom attribute values for the SPARTA grid cell.  The create_particles command can use these per-grid attributes to create particles which (initially) match the continuum flow field solution.  The search for the closest coarse grid point to each SPARTA grid cell is done using a k-d tree in 2 or 3 dimensions.

## Author(s)

Steve

## Backward Compatibility

Syntax of existing custom command was changed somewhat, to make it easier to be compatible with file reading and the new fix custom command

## Implementation Notes

New test scripts added to the examples/custom directory

## Post Submission Checklist


sjplimp:
@stanmoore1 Looks like I still have some work to do to jive old test scripts with the new custom syntax

sjplimp:
@stanmoore1 I think the issue with the old scripts is fixed

stanmoore1:
@sjplimp some of the new examples are failing with: `ERROR: Incorrect line format in custom attribute file `, can you take a look?

stanmoore1 (approved):

PR #554, opened 2025-03-12
Add options to the create_particles command to use custom per-grid attributes to affect properties of the created particles

sjplimp:
## Purpose

The create_particles command has options to use variables that are functions of a created particle's coordinates to alter their properties (temperature, species, streaming velocity, etc).  This enables spatially varying particle initialization.

This PR adds options to use per-grid custom attributes to do something similar.  The per-grid attributes can be read in from files or set by variables.  They are used to set attributes for each grid cell, which are applied when particles in that grid cell are created.  Thus it operates at grid cell resolution, rather than at the resolution of individual particles.

One use for this new feature is useful for creating SPARTA particles which match the flow field produced by a continuum code as output into a file that SPARTA inputs with per-grid attributes.

## Author(s)

Steve

## Backward Compatibility

Should not change existing behavior of other create_particles options.

## Implementation Notes

## Post Submission Checklist

### DIFF
--- a/doc/Section_howto.html
+++ b/doc/Section_howto.html
@@ -1653,13 +1653,21 @@ Each custom attribute has a name, which allows them to be specified in
 input scripts as arguments to various commands.  The values each
 attricute stores can be either integer or floating point numbers.
 </P>
-<P>Here are lists of current commands (as of August 2023) which use custom
-attributes in various ways:
+<P>The <A HREF = "custom.html">custom</A> command can create and set/reset custom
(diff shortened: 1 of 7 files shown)


sjplimp:
@stanmoore1 This feature was requested by Michael.  I still have some testing/debugging to do.  Michael may also want to check it out.

stanmoore1:
@sjplimp can you add an example for the new option?

stanmoore1:
Regression tests passed

stanmoore1 (approved):

PR #556, opened 2025-03-20
New fix lammps command

talinke:
## Purpose

I contribute a concurrent coupling framework between SPARTA and the molecular dynamics solver LAMMPS. The new fix command enables users to provide MD insights to their SPARTA simulation in an automated fashion. The implementation is contained to one new source file. Additions to the build system allow for users to couple LAMMPS through a simple one-argument addition during compilation. The provided documentation clearly outlines the set-up, usage and workflow of the fix. A simple example running for less than 30 seconds highlights its applicability.

## Author(s)

Tim Linke (UC Davis, LLNL)

## Backward Compatibility

N/A

## Implementation Notes

Adds a new fix_lammps.cpp/h command
Adds a new fix lammps doc page
Adds a new examples/couple dir
Adds a new CMake option to enable coupling

## Post Submission Checklist

A video and paper will be provided via links in the documentation shortly.


mpr7366:
With this new release, is it possible to use Sparta per grid quantity as an external force for Lammps Input?

talinke:
Yes, that is certainly possible.

talinke:
A preprint of the method is now available [here](https://doi.org/10.48550/arXiv.2506.01924). A brief video summarizes the capability [here](https://youtu.be/-KR9Ebh1xWY).

talinke:
Thank you for the comments @stanmoore1. I chose my current file protection strategy (see above) since it automatically enables an automatic linking to the LAMMPS library. In my opinion, it serves the same purpose as "#ifdef", but I would be happy to change the implementation.

talinke:
Following your suggestion @stanmoore1, I restructured the implementation as an optional package. The files are now fully protected and support for both CMake and GNU Make is included. The main documentation on building and usage sits in fix_lammps.

stanmoore1:
@talinke I tried building this with GNU Makefile but I'm getting a link error:
```
/usr/bin/ld: math_extra.o (symbol from plugin): in function `MathExtra::write3(double const (*) [3])':
(.text+0x0): multiple definition of `MathExtra::mldivide3(double const (*) [3], double const*, double*)'; math_extra.o (symbol from plugin):(.text+0x0): first defined here
collect2: error: ld returned 1 exit status
```
Any ideas?

stanmoore1:
Same error with CMake. Thank you for making this a package, btw.

stanmoore1:
I'm on RHEL9, GCC 11.5.0, MPICH 4.1.1. Let me know if you need anything else.

stanmoore1:
@talinke this fixes the link issue. Not sure why the other functions don't need `inline` though.
```
diff --git a/src/math_extra.h b/src/math_extra.h
index 33b4a1ba..e008a10a 100644
--- a/src/math_extra.h
+++ b/src/math_extra.h
@@ -75,7 +75,7 @@ namespace MathExtra {
   inline void vecmat(const double *v, const double m[3][3], double *ans);
   inline void scalar_times3(const double f, double m[3][3]);
 
-  int mldivide3(const double mat[3][3], const double *vec, double *ans);
+  inline int mldivide3(const double mat[3][3], const double *vec, double *ans);
   int mldivide4(const double mat[4][4], const double *vec, double *ans);
 
   // quaternion operations
```

talinke:
Great workaround, thank you. And interesting that it is only affecting the mldivide3 function... 
Are you encountering any other issues now?

PR #564, opened 2025-05-05
Add python support via a new python command and python-style variables

sjplimp:
## Purpose

Add Python support to SPARTA via the following new features which enable Python code to be invoked directly by input script commands via SPARTA variables.

(a) new "python-style" variables
(b) a new "python" command which hooks a python-style variable to a Python-coded function.  The Python code can be included in the input script, or in a separate Python file.  The "python" command specifies how the Python function arguments and return value are formatted and coupled to the input script
(c) ability to invoke (a,b) from an equal-style or grid-style variable 

## Author(s)

Steve, but a good chunk of this PR, including the Python invocation code was taken from LAMMPS, where it was written by Axel Kohlmeyer (Temple U) and Richard Berger (LANL).

## Backward Compatibility

N/A

## Implementation Notes

## Post Submission Checklist

### DIFF
--- a/lib/README
+++ b/lib/README
@@ -17,3 +17,5 @@ The libraries in this directory are the following:
 
 kokkos        Kokkos package for GPU and many-core acceleration
                 from Kokkos development team (Sandia)
+python        hooks to the system Python library, used by the PYTHON package
+                from the SPARTA development team
--- a/lib/python/Makefile.sparta
+++ b/lib/python/Makefile.sparta
@@ -1,0 +1,14 @@
+# Settings that the LAMMPS build will import when this package library is used
+# See the README file for more explanation
+ifeq ($(shell type python3 >/dev/null 2>&1; echo $$?), 0)
+PYTHON=python3
+PYTHONCONFIG = python3-config
+else
+PYTHONCONFIG = python-config
+PYTHON=python
+endif
+
+
+python_SYSINC = $(shell which $(PYTHONCONFIG) > /dev/null 2>&1 && $(PYTHONCONFIG) --includes || :)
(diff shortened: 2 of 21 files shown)


sjplimp:
@stanmoore1 Can you try to build this with the new src/PYTHON package installed (e.g. after make yes-python).  I get errors with the compile of python_impl.cpp/h which I do not understand how to fix.  Similar code does not give a problem in LAMMPS.

stanmoore1:
@sjplimp this needs CMake support--I can work on that. Can you add an example?

sjplimp:
@stanmoore1 - yes, please add CMake support.  I added a temporary example in examples/circle/in.circle.pyvar with tmp.py.
It works with no valgrind errors until the very end, when the script exits.  It also gets similar valgrind errors if you run with in.circle, which exercises no Python.  This is if the exe is built with the Python package, though not used in the case of in.circle.  It is used with in.circle.pyvar.  Any ideas ?  Note that the output tmp.grid has correct Python-calculated values in it, so I think this is nearly working.

sjplimp:
@stanmoore1 we also need to add support for SPARTA to parse triple quotes from the input script - in order to have the Python function be in-line in the input script.  Also need new doc pages for all this.

sjplimp:
@stanmoore1 I found the end-of-run bug with the destructors

stanmoore1:
Great, I got CMake working.

stanmoore1:
@sjplimp regression tests are passing now. I think only the doc update is remaining.

sjplimp:
@stanmoore1 This one should be ready for you to review and merge

PR #579, opened 2025-09-10
Bugfix for Kokkos adapt, and remove deprecated code

stanmoore1:
## Purpose

Bugfix for Kokkos adapt, and remove deprecated code

## Author(s)

Stan Moore (SNL), adapt bug reported by Michael Gallis (SNL)

## Backward Compatibility

Yes


stanmoore1:
For reference, this fixes the Kokkos version of `compute lambda/grid` erroring out with:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Deprecation Error: Kokkos::deep_copy extents of views don't match: lambda/grid:temp(701934) ave/grid:vector_grid(705687) 
```

PR #607, opened 2026-02-19
Add KOKKOS support for regions

stanmoore1:
## Purpose

Add KOKKOS support for regions including `block`, `cylinder`, `plane`, and `sphere`.

The `intersect` and `union` region styles are not yet supported by KOKKOS.

## Author(s)

Stan Moore (SNL)

## Backward Compatibility

Yes


stanmoore1:
Regression tests passed.

PR #613, opened 2026-03-20
Treat surface object as a mobile rigid body

sjplimp:
## Purpose

Allow an object defined by a surface triangulation (3d) or line polygon (2d) to be treated as a mobile rigid body which interacts with particles.  This can be useful for modeling the effect of gas flows on a large particulates.

Each timestep, particles which bounce off the object impart force and torque to the rigid body.  These can cause the object to more and rotate in response to the flow.  The flow itself is affected by the presence of the rigid body.  If the object gets too close to other static surfaces (triangles/lines), then a push-off force is applied to the rigid body.

New features (many are still work in progress):

(1) Eix rigid command -- stores attributes of the rigid body, time integrates its motion.
(2) More generalized assignment of moving surfaces to grid cells
(3) Collisions of particles with translating/rotating lines/tris
(4) Push-off forces from other static surfaces
(5) Example input scripts in examples/rigid

## Author(s)

Steve and Ryan McMullen (Sandia), Stan Moore (Sandia), Claude Code

## Backward Compatibility

N/A b/c new feature

## Implementation Notes

See summary above

## Post Submission Checklist


MadanBK1:
Hi Stan and steve,

Could the rigid bodies in this PR be used as DEM-like particles, including gas–particle momentum exchange and contacts between solid particles?

Would it also be possible to add axisymmetric support? I

I have an idea to use this in my plume–surface interaction research.


Thank you,
Madan

stanmoore1:
@MadanBK1

> Could the rigid bodies in this PR be used as DEM-like particles, including gas–particle momentum exchange and contacts between solid particles?

Yes to all. Here are some proof-of-concept movies:


https://github.com/user-attachments/assets/f80349ff-1fd0-4fd2-8cf9-04bba5fb611e

https://github.com/user-attachments/assets/c2fa5fbc-fdf8-4372-83db-19a323cc0210






>Would it also be possible to add axisymmetric support?

I will look into this, thanks for the suggestion.

MadanBK1:
Thank you for the explanation and videos. I noticed that the gas particles leave through the outflow boundaries, while the solid bodies appear to remain inside. Is this due to the boundary settings in these examples, or does the current implementation restrict solid bodies from leaving the domain?

Also, would it be practical to simulate thousands of moving 2D circles representing solid particles, including gas–solid momentum exchange and contacts between them?

stanmoore1:
@MadanBK1 yes rigid bodies can leave the domain too, depending on the BCs. Our goal is to support thousands of interacting rigid bodies, e.g.:

img 01365

I'm still working on the code performance.

PR #621, opened 2026-06-02
Fix wildcard `*` expansion inconsistency for single-value `fix ave/grid` in `dump grid`

### LINKED ISSUE #614: Dump replaces wildcard `*` for single species inconsistently between `compute grid` and `fix ave/grid`
## Summary

I noticed an inconsistency in the `dump grid` command, between dumping a grid-style vector from the `compute grid` and `fix ave/grid` commands.

This shows up in the column labels of the grid dumps if a wildcard asterisk `*` is used when the simulation contains only a single species. The compute dump replaces the asterisk (as expected) with the number 1, but the fix dump leaves the `*` unchanged. 

## Expected behavior

The grid dump file contains columns labelled `c_n[1]` and `f_n[1]`.

## Actual behavior

The grid dump file contains columns labelled `c_n[1]` and `f_n[*]`.

## SPARTA version

The latest version (SPARTA 24Sep2025) on a Mac.

## Minimal Reproducible Example

Input file:

``` 
seed                12345           
dimension           2               
boundary            rr      rr      pp
create_box          0 1     0 1     -0.5 0.5
create_grid         2       2       1
species             species.txt H2O
mixture air         H2O vstream 0.0 0.0 0.0 temp 273.15
create_particles    air n 1000
compute n           grid all species n
fix n               ave/grid all 1 1 10 c_n[*]
dump dg             grid all 10 dump_grid.txt c_n[*] f_n[*] 
run                 10
```

species.txt:

```
CO2             44.0        7.31e-26    2       0.2     8       0.05    959         1.0	        0.0
```

## Additional comments

The species file also contains CO2, so one can quickly add it to the input script. If both H2O and CO2 are in the simulation, then the dump file has, as expected, consistent columns named `c_n[1]`, `c_n[2]`, `f_n[1]` and `f_n[2]`.

stanmoore1:
## Purpose

Fix wildcard `*` expansion inconsistency for single-value `fix ave/grid` in `dump grid`
Fixes #614 

## Author(s)

Stan Moore (SNL) working with Github Copilot (Claude Sonnet 4.6)

## Backward Compatibility

Yes

## Implementation Notes

From Github Copilot (Claude Sonnet 4.6):

`dump grid` inconsistently expands the `[*]` wildcard between `compute grid` and `fix ave/grid` outputs when there is only one species: `c_n[*]` correctly becomes `c_n[1]`, but `f_n[*]` stays as `f_n[*]` in column labels.

**Root cause:** `compute_grid` always sets `size_per_grid_cols = ngroup * nvalue` (≥ 1), while `fix_ave_grid` sets `size_per_grid_cols = 0` when `nvalues == 1` (vector convention). `expand_args` only expands `[*]` when `size_per_grid_cols > 0`, so single-value fixes are silently skipped. Data access happens to be correct because `atoi("*") == 0` maps to vector access — only the column label is wrong.

## Changes

- **`src/input.cpp` — `expand_args`**: For `per_grid_flag` and `per_surf_flag` fix cases, drop the `size > 0` guard. Use `nmax = size_per_grid_cols ? size_per_grid_cols : 1` so that single-value (vector) fixes expand `[*]` → `[1]`.

- **`src/dump_grid.cpp` — `parse_fields`**: When `argindex == 1` but the fix is a per-grid vector (`size_per_grid_cols == 0`), silently normalize `argindex` to `0` (vector access) instead of erroring. This makes `f_n[1]` and `f_n` equivalent for single-value fixes and preserves backward compatibility with existing scripts using bare `f_n`.

- **`src/dump_surf.cpp` — `parse_fields`**: Same normalization for per-surf vector fixes.

### DIFF
--- a/src/dump_grid.cpp
+++ b/src/dump_grid.cpp
@@ -549,8 +549,10 @@ int DumpGrid::parse_fields(int narg, char **arg)
         error->all(FLERR,"Dump grid fix does not compute per-grid info");
       if (argindex[i] == 0 && modify->fix[n]->size_per_grid_cols != 0)
         error->all(FLERR,"Dump grid fix does not calculate a per-grid vector");
-      if (argindex[i] > 0 && modify->fix[n]->size_per_grid_cols == 0)
-        error->all(FLERR,"Dump grid fix does not calculate per-grid array");
+      if (argindex[i] > 0 && modify->fix[n]->size_per_grid_cols == 0) {
+        if (argindex[i] == 1) argindex[i] = 0;
+        else error->all(FLERR,"Dump grid fix does not calculate per-grid array");
+      }
       if (argindex[i] > 0 && argindex[i] > modify->fix[n]->size_per_grid_cols)
         error->all(FLERR,"Dump grid fix array is accessed out-of-range");
 
--- a/src/dump_surf.cpp
+++ b/src/dump_surf.cpp
@@ -605,8 +605,10 @@ int DumpSurf::parse_fields(int narg, char **arg)
         error->all(FLERR,"Dump surf fix does not compute per-surf info");
       if (argindex[i]== 0 && modify->fix[n]->size_per_surf_cols != 0)
         error->all(FLERR,"Dump surf fix does not compute per-surf vector");
-      if (argindex[i] > 0 && modify->fix[n]->size_per_surf_cols == 0)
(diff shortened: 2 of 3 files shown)


stanmoore1:
It turns out this PR is incomplete. A more complete version is found here: https://github.com/stanmoore1/sparta/pull/13. However, this gets into some design/expected behavior issues which I need to discuss with @sjplimp.

PR #625, opened 2026-06-12
Fix per-surf area/weight indexing with multiple surf_react adsorb instances

### LINKED ISSUE #624: `surf_react adsorb` condensation condenses too many particles
### Summary

A condensation reaction defined with `surf_react adsorb` condenses much more particles than the same reaction defined with `surf_react global`. 

### Detailed Description

In the MRE below, there is a specularily reflecting cylinder (the flow region is on the inside). Particles are emitted from the bottom of the cylinder, should collide specularily with the walls and top and then condense/freeze out at the bottom of the cylinder. The bottom of the cylinder (surface group `bottom`) is assigned a `freeze` reaction model with a 50% chance of condensation upon collision. The walls and top of the cylinder (surface group `rest`) are assigned a `freezeNOT` reaction model with a 0% chance of condensation upon collision.

Each reaction model is defined in a `.txt` file to be read in with the `surf_react adsorb` command.
For each reaction model, the next line in the script is a comment with a corresponding `surf_react global` command. If you uncomment this line and comment out the preceding line, then the reaction model is defined from `surf_react global` instead of `surf_react adsorb`.

### Expected Behavior

No matter in which way the condensation probabilities are defined, the same amount of reactions should occur. In particular, all condensation reactions should occur at the bottom of the vent, with the `freeze` reaction.

Output:
```
Surface reaction tallies:
  id freeze style global #-of-reactions 1
    reaction all: 990817
    reaction H2O(g) --> H2O(b): 990817
  id freezeNOT style global #-of-reactions 1
    reaction all: 0
```

### Actual Behavior

Particles condense onto the cylinder walls and top, even though the corresponding probability, defined with `surf_react adsorb`, should be 0. Actually, the majority of condensation occurs at the walls and top.

Output:
```
Surface reaction tallies:
  id freeze style adsorb #-of-reactions 1
    reaction all: 176602
    reaction H2O(g) --> H2O(b): 176602
  id freezeNOT style adsorb #-of-reactions 1
    reaction all: 821068
    reaction H2O(g) --> H2O(b): 821068
```

After commenting the `surf_react freezeNOT adsorb` definition and uncommenting the `surf_react freezeNOT global` definition, the reaction tallies look like we expect.

Output after this change:
```
Surface reaction tallies:
  id freeze style global #-of-reactions 2
    reaction all: 990817
    reaction delete: 990817
  id freezeNOT style global #-of-reactions 2
    reaction all: 0
```

### Additional comments

If the surface group `bottom` is also assigned the surface reaction model `freezeNOT`, the script works as then would be expected, with no particles condensing anywhere.

This is not just a problem with using `0.0` when defining a collision probability. When using e.g. `0.00001` as the collision probability for the `freezeNOT` reactions, `surf_react adsorb` still condenses much more particles than `surf_react global`.

Of course, this is a constructed example, for which it would not be necessary to assign such a reaction model to the walls at all. But in the end I want to simulate multiple species in such a cylinder, with only some of these species condensing onto the walls. For this, assigning species-specific condensation coefficients through text files and `surf_react adsorb` seems necessary.

### SPARTA Version

The latest release (SPARTA 24Sep2025) on a Mac.

### Files

Input file:
```
seed                    12345 
dimension               3              
boundary                oo          oo          oo
create_box              -0.5 0.5    -0.5 0.5    0 1
create_grid             10 10 10 clump xyz

species                 species.txt H2O
mixture air             H2O vstream 0.0 0.0 300.0 temp 300.0

read_surf               cylinder.txt
group bottom            surf id 1:8
group rest              surf subtract all bottom
fix in                  emit/surf air bottom n 1000 perspecies no
surf_collide specular   specular
surf_react freeze       adsorb gs freeze.txt nsync 1 surf 273.15 6.022e18 H2O
#surf_react freeze       global 0.5 0.0
surf_react freezeNOT    adsorb gs freezeNOT.txt nsync 1 surf 273.15 6.022e18 H2O
#surf_react freezeNOT    global 0.0 0.0
surf_modify bottom      collide specular react freeze
surf_modify rest        collide specular react freezeNOT

dump im                 image all 100 image.*.ppm type type surf one 0.005 pdiam 0.01 size 512 512 zoom 1.3 
dump_modify im          pad 4
timestep                1e-3
run                     1000
```

freeze.txt:
```
H2O(g) --> H2O(b)
CD S 0.5 1
```

freezeNOT.txt:
```
H2O(g) --> H2O(b)
CD S 0.0 1
```

species.txt:
```
```

cylinder.txt:
```
#

18 points
32 triangles

Points

1 0.0 0.0 0.0
2 0.5 0.0 0.0
3 0.3535533905932738 0.35355339059327373 0.0
4 3.061616997868383e-17 0.5 0.0
5 -0.35355339059327373 0.3535533905932738 0.0
6 -0.5 6.123233995736766e-17 0.0
7 -0.35355339059327384 -0.35355339059327373 0.0
8 -9.184850993605148e-17 -0.5 0.0
9 0.35355339059327373 -0.35355339059327384 0.0
10 0.5 0.0 1.0
11 0.3535533905932738 0.35355339059327373 1.0
12 3.061616997868383e-17 0.5 1.0
13 -0.35355339059327373 0.3535533905932738 1.0
14 -0.5 6.123233995736766e-17 1.0
15 -0.35355339059327384 -0.35355339059327373 1.0
16 -9.184850993605148e-17 -0.5 1.0
17 0.35355339059327373 -0.35355339059327384 1.0
18 0.0 0.0 1.0

Triangles

1 1 2 3
2 1 3 4
3 1 4 5
4 1 5 6
5 1 6 7
6 1 7 8
7 1 8 9
8 1 9 2
9 18 11 10
10 18 12 11
11 18 13 12
12 18 14 13
13 18 15 14
14 18 16 15
15 18 17 16
16 18 10 17
17 2 17 10
18 2 10 3
19 3 10 11
20 3 11 4
21 4 11 12
22 4 12 5
23 5 12 13
24 5 13 6
25 6 13 14
26 6 14 7
27 7 14 15
28 7 15 8
29 8 15 16
30 8 16 9
31 9 16 17
32 9 17 2
```

stanmoore1:
## Purpose

Fixes #624

## Author(s)

Stan Moore (SNL) and Claude Code (Fable 5)

## Backward Compatibility

Yes

Description from Claude:

## Summary

Fixes the bug reported in [sparta/sparta#624](https://github.com/sparta/sparta/issues/624): with multiple `surf_react adsorb` instances assigned to different surf groups, a CD (condensation) reaction with probability **0.0** still condensed large numbers of particles, while the equivalent `surf_react global 0.0` behaved correctly.

## Root cause

In `SurfReactAdsorb::init()`, the one-time initialization of the custom per-surf `area` and `weight` values (non-distributed surfs branch) used an index `m` that was only incremented for surf elements assigned to *this* reaction model instance:

```cpp
for (int isurf = me; isurf < nslocal; isurf += nprocs) {
  isr = tris[isurf].isr;
  if (surf->sr[isr] != this) continue;   // skips m++
  area[m] = surf->tri_size(&tris[isurf],tmp);
  weight[m] = 1.0;
  m++;
}
```

`m` is an index into the *owned-surf* custom arrays (one entry per owned surf), so it must advance for every owned surf — as the distributed branch already does by indexing with `isurf` directly. With two or more adsorb instances, each instance instead wrote its values compacted at the front of the shared custom arrays:

- values landed on the wrong surf elements,
- instances overwrote each other's values, and
- trailing entries were never written and stayed at their initialized value of 0.

For a surf left with `area = weight = 0`, `react()` computes `factor = fnum * weight / area = 0/0 = NaN`, which propagates through `surf_cover`, `S_theta`, and `prob_value` (`0.0 * NaN = NaN`). Every comparison against NaN is false, so `react_prob > random_prob` fails and the `react_prob <= random_prob → continue` guard also fails — the reaction **always executes** on those surfs, regardless of its probability coefficient. This explains all the reported symptoms: spurious condensation on surfs with probability 0.0, persisting even with tiny probabilities, and disappearing when a single model is assigned to all surfs (compaction is then harmless).

## Fix

Increment `m` on every owned surf and only write `area`/`weight` when the surf belongs to this instance, matching the distributed-surfs branch.

## Verification

Minimal 2D repro (circle split into two groups, two `adsorb gs` instances: `CD S 0.5` on half the surfs, `CD S 0.0` on the rest, 500 steps):

| | prob 0.5 model | prob 0.0 model |
|---|---|---|
| before fix | 7270 reactions | **13264 reactions** |
| after fix | 7245 reactions | **0 reactions** |

Regression check: `examples/surf_react_adsorb/in.circle.gs` (single instance assigned to all surfs) produces reaction tallies identical to the shipped reference log `log.11Sep23.mpi_1.circle.gs` (26588 total, per-reaction counts all match).

### DIFF
--- a/src/surf_react_adsorb.cpp
+++ b/src/surf_react_adsorb.cpp
@@ -543,9 +543,10 @@ void SurfReactAdsorb::init()
       int m = 0;
       for (int isurf = me; isurf < nslocal; isurf += nprocs) {
 	isr = lines[isurf].isr;
-	if (surf->sr[isr] != this) continue;
-	area[m] = surf->line_size(&lines[isurf]);
-	weight[m] = 1.0;
+	if (surf->sr[isr] == this) {
+	  area[m] = surf->line_size(&lines[isurf]);
+	  weight[m] = 1.0;
+	}
 	m++;
       }
     } else {
@@ -553,9 +554,10 @@ void SurfReactAdsorb::init()
       int m = 0;
       for (int isurf = me; isurf < nslocal; isurf += nprocs) {
 	isr = tris[isurf].isr;
-	if (surf->sr[isr] != this) continue;
-	area[m] = surf->tri_size(&tris[isurf],tmp);
-	weight[m] = 1.0;
+	if (surf->sr[isr] == this) {
+	  area[m] = surf->tri_size(&tris[isurf],tmp);
+	  weight[m] = 1.0;
+	}
 	m++;
       }
     }


stanmoore1:
From Claude:

If you prefer an explicit mapping over a running counter, the equivalent formulation (used by fix_surf_temp.cpp:232) is:

```c++
for (int i = 0; i < nsown; i++) {        // i = owned index
  int isurf = me + i*nprocs;             // local index
  if (surf->sr[lines[isurf].isr] != this) continue;
  area[i] = surf->line_size(&lines[isurf]);
  weight[i] = 1.0;
}
```
That's arguably clearer because the continue is now harmless — the index i is loop-controlled rather than manually maintained. I kept the m++ form in the PR only to stay minimal and match the surrounding code's style; happy to switch it to the explicit-mapping form if you'd rather have that on the branch.

PR #627, opened 2026-06-17
Small bugfixes

### LINKED ISSUE #349: SPARTA compilation with Kokkos
## Summary

SPARTA Compilation with Kokkos

## Type of Issue

Help Request?

I am working on compiling SPARTA with the kokkos backend provided in the January 7, 2022 release. When using cmake to build kokkos in the lib/kokkos directory, i've noted that libkokkoscontainers.so and libkokkoscore.so are built, but not libkokkos.so as seems to be expected by SPARTA. Is there anything special i have to include while building Kokkos?

No error that lkokkos is not found on compilation.

Thanks in advance!

### LINKED ISSUE #619: compute react/surf per-column tally appears off-by-one for surf_react adsorb PS reactions
Hello,                                                                                                                                                                                
                                                                            
  Thank you for SPARTA - it has been central to our coupled DSMC +                                                                                                                      
  finite-volume ablation work.                                                                 
                                                                                           
  While instrumenting a case with `compute react/surf` attached to a                                                                                                                    
  `surf_react adsorb gs/ps` model, we noticed that the on-surface (PS)    
  reactions in the per-column dump appear to land one column lower than                                                                                                                 
  file order, and the column reserved for the last PS reaction reads                           
  zero in every run we tried. The GS columns look correct.                                                                                                                              
                                         
[sparta_ps_offbyone_bug_report_2026-05-29.zip](https://github.com/user-attachments/files/28412645/sparta_ps_offbyone_bug_report_2026-05-29.zip)
                                                                                                                                               
  Tracing the code, our reading is that `surf_tally` expects a                                                                                                                          
  1-indexed reaction argument, and the GS path follows this via                                                                                                                         
  `return list[i] + 1` - but the PS path in `SurfReactAdsorb::PS_react`                                                                                                                 
  (near line 2950 of upstream master) passes the raw `ireaction` value
  without the `+1`. We may well be misreading the design.                                                                                                                               
                                                                                                                                                                                        
  A small self-contained bundle is attached with:                                                                                                                                       
  - a short write-up of what we saw                                                                                                                                                     
  - a drop-in reproducer using your bundled                                                                                                                                             
    `examples/surf_react_adsorb/in.beam.surf.gs_ps` (3 added lines)                                                                                                                     
  - the empirical dump outputs before and after a one-character change                                                                                                                  
    we tried                                                                                                                                                                            
  - the one-line change in unified-diff form                                                                                                                                            
  - a script that runs the before/after demo against any SPARTA tree                                                                                                                    
    (~5 minutes)                                                                                                                                                                        
                                                                                                                                                                                        
  Verified against upstream master `1760f614`.                                                                                                                                          
                                                                     
                                                                                                                         
                                                                                                                                                                            
  If we have misread something, we'd be grateful to know. Happy to open                                                                                                                 
  a PR with the change, plus a small regression test if useful.                                                                                                                          
                                                                                                                                                                                        
  Thank you again for SPARTA.

### LINKED ISSUE #626: custom surf set on a surface group writes values to the wrong surfaces
Hi everyone,

I’ve been looking into some unexpected behavior with the custom surf set <attr> <var> <group> command, and I suspect there might be a bug in how it selects surfaces. It appears to write values to the lowest-indexed surfaces instead of those in the specified <group>.

While the correct number of surfaces are updated, they seem to be the wrong ones. This happens silently with no error, though I’ve noticed that setting on group all works as expected.

Could you help me verify if this is indeed a bug? I’ve compiled the following details for review:

Scope
- Affected Version: 24Sep2025 (also observed in latest master 5aed836e).

Investigation Findings
It looks like the issue may be in src/custom.cpp within Custom::set_surf (~line 991). It appears to use choose[count++], while set_particle and set_grid use choose[i]. It seems like a potential one-line fix, but I would appreciate a second pair of eyes on this.

Verification Details
I have attached a folder containing the following for testing:
- 01_issue_body.md: Breakdown of the observed vs. expected behavior.
- 03_reproducer/: Minimal runnable reproducer (in.repro + mini.surf).
- 04_outputs/: Dumps showing the behavior before/after.
- 06_patch/: The suspected one-line patch.
- run_verification.sh: A script that builds the patched copy and verifies the output (checked in serial, -np 2, 2D, and 3D).

Please let me know your thoughts—thanks for your time!

[sparta_custom_surf_set_subgroup_bug_report_2026-06-09.zip](https://github.com/user-attachments/files/28906852/sparta_custom_surf_set_subgroup_bug_report_2026-06-09.zip)

stanmoore1:
## Purpose

Small bugfixes:

Fixes #626 
Fixes #619 
Fixes #349 

## Author(s)

Stan Moore (SNL) with help from AI (ChatGPT 5.5, Claude Code Opus 4.8), and Upendra Yadav (University of California, Irvine)

## Backward Compatibility

Yes

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -103,6 +103,8 @@ set(TARGET_SPARTA spa_${SPARTA_MACHINE})
 
 set(TARGET_SPARTA_LIB sparta_${SPARTA_MACHINE})
 if(BUILD_SHARED_LIBS)
+  # Force PIC so intermediate static package libs link into the shared libsparta
+  set(CMAKE_POSITION_INDEPENDENT_CODE ON)
   set(TARGET_SPARTA_LIB_LINK "libsparta${CMAKE_SHARED_LIBRARY_SUFFIX}")
   set(TARGET_SPARTA_LIB_NAME
       "libsparta_${SPARTA_MACHINE}${CMAKE_SHARED_LIBRARY_SUFFIX}")
--- a/src/custom.cpp
+++ b/src/custom.cpp
@@ -56,7 +56,7 @@ Custom::~Custom()
 
   for (int i = 0; i < naction; i++) {
     int action = actions[i].action;
-    if (action == FILESTYLE) {
+    if (action == FILESTYLE || action == FILECOARSE) {
       delete [] actions[i].fname;
       delete [] actions[i].cindex_file;
       delete [] actions[i].ctype_file;
@@ -431,9 +431,9 @@ bigint Custom::process_actions(int narg, char **arg, int external)
       // # of coarse files and filestyle
 
       int numfile = input->inumeric(FLERR,arg[iarg+1]);
-      int filestyle;
       if (strcmp(arg[iarg+2],"text") == 0) filestyle = TEXT;
       else if (strcmp(arg[iarg+2],"binary") == 0) filestyle = BINARY;
+      else error->all(FLERR,"Illegal custom command");
 
       // file name
       // if numfile > 1, fname must have "%" wildcard char
@@ -650,6 +650,9 @@ bigint Custom::process_actions()
       // assign grid cell attributes from ave of neighbor coarse point values
 
       count += coarse_tree_neighbor_assign(1,colcount,cindex,ctype,csize,ccol);
+
+      memory->destroy(xyz_coarse);
+      memory->destroy(values_coarse);
 
       // for mode = GRID
       //   set estatus of all changed custom vecs/arrays to 0
@@ -963,8 +966,10 @@ bigint Custom::set_surf(int groupbit, Region *region,
   int flag;
   double point[3];
 
+  int m = 0;
+
   bigint count = 0;
-  for (int i = start ; i < stop; i += skip) {
+  for (int i = start ; i < stop; i += skip, m++) {
     flag = 1;
     if (dim == 2) {
       if (!(lines[i].mask & groupbit)) flag = 0;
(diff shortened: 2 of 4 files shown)


sjplimp (approved):

UPENDRAYADAV07:
Hi @stanmoore1 ,

Thank you for merging this PR! I wanted to politely mention that I contributed to some of the fixes in this work.

I notice that:
1. I'm already credited as a co-author on the "Fix off-by-one bug" commit in this PR ✓
2. However, I also contributed to the "Fix various issues in custom" commit (a634ee93) which fixed issue #626, but I'm not listed as a co-author there.

Additionally, I wanted to mention that I also contributed to commit c4f33ab4 which fixed issue #641 (the stale per-surf temperature pointer segfault), but I'm not included as a co-author on that commit either.

Would it be possible to add co-author trailers to both of these commits to reflect my contributions? This would help ensure the contribution history is accurate.

Thank you for considering this!

stanmoore1:
@UPENDRAYADAV07 the reason I didn't co-author https://github.com/sparta/sparta/commit/a634ee93be392f6d4f5d23942c5891538ca91665 and https://github.com/sparta/sparta/commit/c4f33ab47d40f2e985f06ca2c211cb3344cff05f is because the bug fixes ended up being different than you proposed, but your suggestions were certainly helpful, so I'm happy to amend the commits to co-author. Give me a bit.

UPENDRAYADAV07:
Thank you so much @stanmoore1 for the clarification and for being willing to amend the commits! I really appreciate that.

I completely understand that implementations often evolve as bugs are debugged and solutions are refined. Your explanation makes total sense, and I'm grateful that my suggestions were helpful in pointing you in the right direction, even if the final fixes took a different approach.

I'm happy with whatever you decide works best. Adding the co-author credits would be great, but I'm not concerned either way. Thanks again for the collaboration and for taking the time to explain the thought process behind these fixes!

stanmoore1:
@UPENDRAYADAV07 this is fixed in https://github.com/sparta/sparta/pull/649 --`master` branch is protected and force pushing is not allowed in this case, so I reverted and reapplied the commits with the co-author trailer.

UPENDRAYADAV07:
@stanmoore1  Thank you very much.

PR #628, opened 2026-06-18
Port subsonic emission in `fix emit/face/kk` to use Kokkos

### LINKED ISSUE #574: Kokkos porting of fix emit/face subsonic
# Summary

When using fix emit/face subsonic for pressure boundary condition it is not yet possible to utilize kokkos acceleration. Porting of this under _kk would make it possible to accelerate e.g. on GPU.

# Type of issue 

Suggestion for an Enhancement

stanmoore1:
## Purpose

Port subsonic emission in `fix emit/face/kk` to use Kokkos
Fixes sparta/sparta#574

## Author(s)

## Backward Compatibility

Yes 

## Implementation Notes

Description from Claude

fix emit/face/kk aborted with "Cannot yet use fix emit/face/kk with
subsonic emission". Port the subsonic pre-processing to run on the
device with Kokkos parallel dispatch instead of falling back to the
host code:

- subsonic_sort() uses ParticleKokkos::sort_kokkos() to build the
  compressed per-cell particle lists (d_plist/d_cellcount) on the
  device, in lieu of the host linked lists

- subsonic_grid() computes per-task nrho, vstream, thermal/rot/vib
  temperatures (and per-species vscale for PONLY) in a parallel_for
  over tasks, accumulating particle moments from the sorted cell
  lists. The TEMPLIMIT check uses a device atomic_max reduction so
  the existing subsonic_temperature_check() warning is preserved

- subsonic_inflow() recomputes per-task ntarget/ntargetsp in a
  parallel_reduce over tasks using a device version of
  FixEmit::mol_inflow(), flagging insertion counts that exceed a
  32-bit int

The subsonic methods are made virtual in FixEmitFace so the Kokkos
class can override them. Task data is modified in place on the device
(k_tasks/k_ntargetsp/k_vscale marked modified on device), so no
host-device transfers of particle or task data are needed during
timestepping.

Also refresh particle_kk_copy before the subsonic kernels since the
particle data structures may have been reallocated (sort/grow) since
the copy taken in the previous perform_task() call; dispatching with
a stale copy corrupts Kokkos view reference counts. Also fix the
prefactor member (used by the modulate keyword) being truncated from
double to int.

Both PTBOTH (subsonic P T) and PONLY (subsonic P NULL) styles agree
with the CPU implementation on the circle example with perspecies
yes/no, and runs are valgrind-clean.

### DIFF
--- a/src/KOKKOS/Install.sh
+++ b/src/KOKKOS/Install.sh
@@ -141,6 +141,7 @@ action create_particles_kokkos.cpp
 action create_particles_kokkos.h
 action fix_emit_face_kokkos.cpp
 action fix_emit_face_kokkos.h
+action fix_emit_kokkos.h
 action fix_emit_surf_kokkos.cpp
 action fix_emit_surf_kokkos.h
 action fix_grid_check_kokkos.cpp
--- a/src/KOKKOS/fix_emit_face_kokkos.cpp
+++ b/src/KOKKOS/fix_emit_face_kokkos.cpp
@@ -32,6 +32,7 @@
 #include "error.h"
 #include "kokkos_type.h"
 #include "particle_kokkos.h"
+#include "grid_kokkos.h"
 #include "sparta_masks.h"
 #include "variable.h"
 #include "Kokkos_Random.hpp"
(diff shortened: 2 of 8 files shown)


stanmoore1:
Subsumed into https://github.com/sparta/sparta/pull/632

PR #630, opened 2026-06-18
Fixed type of parameter passed to sparta_extract_compute in Python wrapper.

garlib:
Fixed wrong parameter type. Parameter style should be an integer, but had been passed as encoded string according to function signature in library.h.

## Purpose

_Python wrapper implementation encode the parameter style as utf8 string leading to error on calling C library function sparta_extract_compute. Checking possible calls of said function in Python wrapper and comparison with function signature in library.h concludes to passing integer type instead._

## Author(s)

Daniel G.

## Backward Compatibility

No, as parameter type changed  and existing code might relied on string-based styles.

## Implementation Notes

Removed code to convert parameter before passing. Type check and enforced conversion to integer might be of use.

## Post Submission Checklist

Please check the fields below as they are completed

None.

## Changes

**python/sparta.py**: Removed 
```
style = style.encode('utf-8')
```

### DIFF
--- a/python/sparta.py
+++ b/python/sparta.py
@@ -74,7 +74,6 @@ class sparta:
     return ptr[0]
 
   def extract_compute(self,id,style,type):
-    style = style.encode('utf-8')
     if type == 0:
       self.lib.sparta_extract_compute.restype = POINTER(c_double)
       ptr = self.lib.sparta_extract_compute(self.spa,id,style,type)


stanmoore1:
@garlib thank you for reporting this bug, and the pull request. I ran this through an AI, and it found a few more issues so I'm closing this PR in favor of https://github.com/sparta/sparta/pull/636. Please let us know if you have any more issues.

PR #638, opened 2026-06-26
Fix vibrational/rotational energy sampling cutoff for many DOF (issue #64)

### LINKED ISSUE #64: Random vibrational energy generation
## Summary

In the function Particle::evib in particle.cpp, when the number of vibrational degrees (vibdof) is greater than 2, there is a hardcode energy cut-off of 10 kT in the vibrational energy sampling. This cut-off is too small when vibdof is large.

## Type of Issue

Bug Report

If one considers a molecule with a rather high number of vibrational dof (in my case SF6 which has vibdof=15) and vibstyle == SMOOTH, the hardcode cut-off of 10 kT is too small and the vibrational energy sampling is not good. I suggest coding a cut-off which depends on vibdof, like for example replacing line 1061 with:
        erm = 20.0*a*erandom->uniform();

Current git.

stanmoore1:
## Purpose

Fix vibrational/rotational energy sampling cutoff for many DOF
Fixes https://github.com/sparta/sparta/issues/64

## Author(s)

## Backward Compatibility

Trajectories are slightly changed, needed to rebless the vibrate example logs.

## Implementation Notes

Description from Claude:

## Summary

Fixes #64 ("Random vibrational energy generation").

`Particle::evib()` (and the structurally identical `Particle::erot()`) sample a particle's internal energy from the equilibrium distribution

```
P(x) ∝ x^a · exp(−x),   x = E/kT,   a = dof/2 − 1
```

using acceptance–rejection. The acceptance function `b = (x/a)^a · exp(a−x)` is correct, but the **candidate energies were drawn uniformly on a fixed range `[0, 10 kT]`**:

```cpp
erm = 10.0*erandom->uniform();
```

This distribution is `Gamma(shape = a+1)`, with **mean = a+1 = dof/2**. The fixed `10 kT` ceiling is only safe while the mean sits well below 10. As the number of degrees of freedom grows, the cutoff falls below the mean of the distribution, so the sampler can never propose the energies where most of the probability mass lives. The result is vibrational energy that is **biased low and whose distribution shape is destroyed** (detailed balance is violated). For SF₆-class polyatomics (~30 vibrational DOF) the bias is severe.

This affects every code path that initializes vibrational energy at a temperature — inflow/emit boundaries (`fix_emit_*`) and diffuse/CLL/TD surface models (`surf_collide_*`).

## Fix

Scale the candidate range with the number of degrees of freedom so it always covers the high-energy tail. The mode is at `x = a`, the mean at `a+1`, the std dev `√(a+1)`; using **mean + ~9 std devs**:

```cpp
double xmax = a + 1.0 + 9.0*sqrt(a+1.0);
erm = xmax*erandom->uniform();
```

This keeps the rejection efficiency roughly constant (~18–20%) instead of letting it collapse, and reduces to the original behavior for small molecules. The rotational path is bounded (`rotdof ≤ 3`, so `a ≤ 0.5`), meaning the bug was latent there, but the same fix is applied for consistency.

Changed file: `src/particle.cpp` (`evib()` and `erot()`).

## Verification

A standalone Monte-Carlo test reproduces the **old** and **new** samplers verbatim, drawing 2×10⁶ samples per case, and compares the sampled mean and variance of `E/kT` against the analytic values (`mean = variance = vibdof/2`).

| vibdof | OLD mean (err) | OLD var | NEW mean (err) | NEW var | theory | OLD eff | NEW eff |
|--------|----------------|---------|----------------|---------|--------|---------|---------|
| 4  | 2.00 (−0.07%) | 1.97 | 2.00 (−0.05%) | 1.99 | 2.0  | 27% | 18% |
| 8  | 3.92 (−1.9%)  | 3.46 | 4.00 (+0.07%) | 4.00 | 4.0  | 44% | 20% |
| 16 | 6.84 (**−14.5%**) | 3.20 | 8.00 (−0.04%) | 7.99 | 8.0  | 52% | 20% |
| 24 | 8.25 (**−31.3%**) | 1.68 | 12.00 (−0.04%) | 11.99 | 12.0 | 25% | 19% |
| 30 | 8.76 (**−41.6%**) | 1.02 | 15.00 (−0.03%) | 15.00 | 15.0 | 8%  | 19% |
| 40 | 9.20 (**−54.0%**) | 0.51 | 20.00 (−0.02%) | 20.01 | 20.0 | 0.4% | 18% |

Findings:

1. **The bug is real and severe** — the old mean collapses (−42% at SF₆-class `vibdof=30`, −54% at `vibdof=40`) and asymptotes toward ~9–10 because returned energies are clamped to `[0, 10 kT]`. The variance collapses even harder, confirming the *shape* is wrong, not just the mean.
2. **The fix recovers both moments** to within ~0.07% of theory across the whole range, so the full distribution is restored.
3. **Efficiency is improved at high DOF** — the old sampler's acceptance rate drops to 0.4% at `vibdof=40` (its `[0,10]` window rarely lands near the mode at `a=19`), whereas the new sampler holds ~18–20% throughout.
4. **Small molecules are unchanged** (`vibdof = 4, 8`), confirming the fix doesn't perturb cases that already worked.

The analytic percent-error curves agree exactly with the Monte-Carlo points:

image

### DIFF
--- a/examples/vibrate/log.07Jul26.mpi_1.vibrate
+++ b/examples/vibrate/log.07Jul26.mpi_1.vibrate
@@ -1,4 +1,4 @@
-SPARTA (7 Mar 2024)
+SPARTA (24 Sep 2025)
 Running on 1 MPI task(s)
 ################################################################################
 # test of vibrational energy modes
@@ -22,8 +22,8 @@ create_box  	    0 1e-5 0 1e-5 0 1e-5
 Created orthogonal box = (0 0 0) to (1e-05 1e-05 1e-05)
 create_grid 	    2 2 2
 Created 8 child grid cells
-  CPU time = 0.000811451 secs
-  create/ghost percent = 98.6133 1.38665
+  CPU time = 0.000814469 secs
+  create/ghost percent = 97.2833 2.71674
 
 species		    co2.species N2 CO2 vibfile co2.species.vib
 
@@ -39,7 +39,7 @@ fix                 1 vibmode
 
 create_particles    mix n 0 twopass
 Created 1000 particles
-  CPU time = 0.00164516 secs
+  CPU time = 0.00144881 secs
 
 variable            collrate equal "ncoll*step*2/np"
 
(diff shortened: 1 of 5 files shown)


stanmoore1:
@aborner1 @mgallis @sjplimp since this is a physics issue and changes the numerics slightly, I'd really like your review before merging this.

aborner1:
@stanmoore1 This is elegant and simple enough. I've looked through the code, and given how little intrusive it is and few changes there are, I am good with it from just a visual inspection.

My comment is actually not related to the changes in PR, but something I noticed when looking at the files and is grandfathered in.
I know that rotstyle=DISCRETE is not yet turned on, but features are already coded in a lot of places. What I am noticing is that in the erot functions in all the files that were edited here, there is an "else" statement that would catch any particle with rotdof > 2 for both DISCRETE and SMOOTH. Is that the intent? Or would there be additional code added in something like "fix_rotmode" similarly to what's happening with vibrational modes?
On that note, should AI agents finish implementing the rotstyle=DISCRETE option?

sjplimp (approved):

PR #642, opened 2026-06-29
Allow ambipolar ions as third bodies in dissociation/recombination (issue #176)

### LINKED ISSUE #176: Using ambipolar species as third bodies in dissociation or recombination reactions
## Summary

Turn on the ability to use ambipolar species third body in dissociation or recombination reaction.

## Type of Issue

Suggestion for an Enhancement

With the new ambipolar reaction patch added, it seems like ions should be able to be used as third bodies for both dissociation and recombination reactions. However, the code does not allow that (returns an error message). It seems like the logic would simply need to be fixed to allow those reactions.

e.g.:
D: O2 + O2+ --> O + O + O2+
R: O + O --> O2 + O2+

stanmoore1:
## Purpose

Allow ambipolar ions as third bodies in dissociation/recombination
Implements and therefore closes #176 

## Author(s)

## Backward Compatibility

Yes

## Implementation Notes

Description from Claude

When the ambipolar approximation is enabled, SPARTA validated every reaction
involving an ambipolar ion or electron against a fixed whitelist of formats in
`ReactBird::ambi_check()`. That whitelist had no entry for an ambipolar **ion
acting as an inert third body** (collision partner) in otherwise-neutral
chemistry, so reactions such as

```
D: O2 + O2+ -> O + O2+ + O
R: O + O  -> O2 + O2+
```

were rejected at setup with `ERROR: Invalid ambipolar reaction`. Physically an
ion is a perfectly valid third body for dissociation and three-body
recombination (charged third bodies often dominate recombination in plasmas),
so this was a real modeling limitation rather than a physics constraint. This
PR allows ambipolar ions to serve as inert third bodies in dissociation and
recombination reactions.

- **`react_bird.cpp` (`ambi_check`)**: accept dissociation with an ion third
  body (for both neutral and ionic dissociating species) and recombination with
  an ion third body. The ion third body must be written as the **middle**
  dissociation product (so it maps to the unaltered second colliding particle in
  `ReactTCE::attempt()`, matching the existing electron-third-body convention),
  or as the **2nd recombination product P2** (which only selects the rate; the
  third body is a spectator drawn from the cell, never a colliding particle).

- **`collide.cpp` (`ambi_reset`)**: for reactions with no electron
  (`jsp != e`), sync each surviving particle's `ionambi` flag to its
  post-reaction species. This is required because, unlike the electron, an ion
  third body is **not** force-swapped into the J slot, so the I/J reactant order
  is not canonical and the positional product mapping could otherwise leave
  `ionambi` inconsistent with the species. The sync is a no-op for all-neutral
  reactions (flags stay 0) and also repairs the analogous latent case for ionic
  recombination products. `fix ambipolar`'s `ions[]` vector is now stored on
  `Collide` for this lookup.

- **Electron-conservation safety**: ion third-body reactions create or destroy
  no electrons, so the per-cell `ion-count == electron-count` invariant is
  preserved; `velambi` is reassigned per-cell from the electron pool regardless.

- **Kokkos**: `ambi_reset_kokkos` mirrors the host logic and a `d_ions` device
  view is wired up from `FixAmbipolarKokkos`.

- **Docs**: `doc/react.txt` documents the new dissociation/recombination
  orderings and the middle-product / P2 requirement.

**Verification performed:**
- Built and ran CPU (serial + MPI), KOKKOS Serial with `SPARTA_KOKKOS_EXACT`,
  and KOKKOS OpenMP.
- **No regression**: A/B comparison against a baseline binary built from the
  parent commit shows *identical* stats output on `examples/ambi`; the baseline
  rejects the new reactions while the patched build accepts them.
- **CPU vs Kokkos**: KOKKOS Serial+EXACT matches the CPU build bit-for-bit
  (all stats columns) on the new example; KOKKOS with 4 MPI ranks matches CPU
  with 4 MPI ranks bit-for-bit.
- **Threads/MPI**: clean runs that conserve electrons on KOKKOS OpenMP (4
  threads), CPU (4 MPI ranks), and KOKKOS (4 MPI ranks).
- **Functional correctness**: in the new example the ion third body (O2+) count
  stays exactly constant at every timestep while O2 dissociates into O,
  confirming the ion behaves as a true inert spectator; no
  `Invalid ambipolar reaction` or electron-count errors in any configuration.

New example `examples/ambi_3body` is a single-group, periodic box of hot
O2 / O / O2+ whose only reactions are an ion-third-body dissociation and
recombination, so any gas reactions that occur exercise the new functionality
directly. A single-group mixture is used because the KOKKOS package currently
supports only single-group collisions (a pre-existing limitation, unrelated to
this change, which is why the multigroup `examples/ambi` case does not run on
Kokkos).

### DIFF
--- a/doc/react.txt
+++ b/doc/react.txt
@@ -290,7 +290,14 @@ Ambipolar dissociation reactions must list their reactants and
 products in one of the following orders:
 
 AB + e -> A + e + B
-AB+ + e -> A+ + e + B :pre
+AB+ + e -> A+ + e + B
+AB + C+ -> A + C+ + B
+AB+ + C+ -> A+ + C+ + B :pre
+
+In the last two forms an ambipolar ion C+ acts as an inert third
+body (collision partner).  As with the electron "e", the third body
+must be written as the middle product, so that it maps to the
+unaltered second colliding particle.
 
 Ambipolar ionization reactions with 2 or 3 products must be in one of
 the following orders:
@@ -308,7 +315,14 @@ Ambipolar recombination reactions must be in the following order:
 
 A+ + e -> A
 A + B+ -> AB+
-A+ + B -> AB+ :pre
+A+ + B -> AB+
+A + B -> AB + C+ :pre
+
+In the last form an ambipolar ion C+ acts as the inert third body,
+written as the second product P2 (see the non-ambipolar discussion of
+3rd particles above).  The two reactants and the recombined molecule
+P1 must be neutral; the ion C+ is a spectator whose species selects the
+recombination rate.
 
 A third particle for recombination reactions can be specified in
 the same way as described above for non-ambipolar recombination.
--- a/examples/ambi_3body/ambi_3body.species
+++ b/examples/ambi_3body/ambi_3body.species
@@ -1,0 +1,17 @@
+# Species data for ambipolar third-body test (subset of air.species)
+
+# ID
+# Molwt (amu)
+# Molmass (kg)
+# Rotational dof
+# RotRel
+# Vibrational dof
+# VibRel
+# VibTemp (K)
+# species wt
+# charge
+
+O2  32.00    5.31E-26  2    0.2   2    5.58659E-5    2256.0    1.0      0.0
+O   16.00    2.65E-26  0    0.0   0    0.0           0.0       1.0      0.0
+O2+ 32.00    5.31E-26  2    0.2   2    5.58659E-5    2256.0    1.0      1.0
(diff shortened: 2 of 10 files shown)


stanmoore1:
@aborner1 can you please review/test this new feature?

aborner1:
> @aborner1 can you please review/test this new feature?

Looks good to me. Tested the non-Kokkos version and no issue. Ran new example as well. Can be merged

stanmoore1:
Thanks @aborner1!

PR #653, opened 2026-07-09
Flag invalid use of wildcard character in compute/fix/etc args used as inputs by various commands

### LINKED ISSUE #614: Dump replaces wildcard `*` for single species inconsistently between `compute grid` and `fix ave/grid`

sjplimp:
## Purpose

Dump and other commands allow input args with a wildcard "*" to expand to represent one or more inputs.  This was allowed in some cases even when the input arg represented a single value, e.g. a per-grid vector and not a per-grid array.  In the case of dump commands this means the input arg was echoed into the dump file column headers with the wildcard, which is confusing.  The solution is to not allow use of a wildcard when it does not make sense, e.g. when a per-grid vector is referenced,

This PR enforces that convention.  It resolves issue #614.

Fixes #614 

## Author(s)

Steve

## Backward Compatibility

Enforcing this convention could trigger errors for input scripts which are using a wildcard incorrectly, but otherwise ran fine.  I.e. they produced a dump file with incorrect column header names, but the simulation and data in the dump file was correct.

## Implementation Notes

## Post Submission Checklist

### DIFF
--- a/examples/ablation/in.ablation.multi.inner.3d
+++ b/examples/ablation/in.ablation.multi.inner.3d
@@ -50,7 +50,7 @@ create_isurf            all fablate 100.5 multi
 timestep		0.001
 
 compute 1 property/grid all vol
-compute 1r reduce sum c_1[*]
+compute 1r reduce sum c_1
 
 #variable vout equal c_1r
 #fix massr print 5 "${vout}" append vol_inner_multi.dat screen no
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -637,7 +637,6 @@ int Input::expand_args(int narg, char **arg, int mode, char **&earg)
         if (ptr2) {
           *ptr2 = '\0';
           if (strchr(ptr1,'*')) {
-            nmax = 0;
 
             // compute
 
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -637,6 +637,7 @@ int Input::expand_args(int narg, char **arg, int mode, char **&earg)
         if (ptr2) {
           *ptr2 = '\0';
           if (strchr(ptr1,'*')) {
+            nmax = 0;
 
             // compute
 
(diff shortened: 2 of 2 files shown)


stanmoore1:
Thanks @sjplimp!

stanmoore1 (approved):

PR #668, opened 2026-08-13
Many collected small changes and bugfixes

### LINKED ISSUE #492: Negative Natt, possibly due to overflow
## Summary

When the number of attempted collisions (Natt in the stats output) grows very large, it rolls over to a negative value. 

## Type of Issue

Bug Report

7 Mar 2024

Natt should always be positive. If the present value exceeds the internal type's maximum value, I would think Sparta should throw an error (or move to a larger type) rather than showing negative values.

The output below is from the `in.collide` script, with the following changes:

```
global              nrho 7.07043E24
global              fnum 7.07043E9
...
#create_particles    air n 10000 twopass
fix in emit/face air all perspecies no n 10000 nevery 1
...
timestep 	    7.00E-7
run 		    10
```

```
SPARTA (7 Mar 2024)
Running on 1 MPI task(s)
Created orthogonal box = (0 0 0) to (0.0001 0.0001 0.0001)
Created 1000 child grid cells
  CPU time = 0.00133896 secs
  create/ghost percent = 75.0534 24.9466
Balance grid migrated 0 cells
  CPU time = 0.000365973 secs
  reassign/sort/migrate/ghost percent = 36.9381 0.521173 10.3583 52.1824
Memory usage per proc in Mbytes:
  particles (ave,min,max) = 0 0 0
  grid      (ave,min,max) = 1.51379 1.51379 1.51379
  surf      (ave,min,max) = 0 0 0
  total     (ave,min,max) = 1.51379 1.51379 1.51379
Step CPU Np Natt Ncoll c_temp 
       0            0        0        0        0            0 
       1     11.73383    10000 88116753 59829525    366.63407 
       2    68.489725    20000 437572007 279701457    367.18408 
       3    204.56655    30000 1058211711 656481202    365.11625 
       4    450.31007    40000 1932325651 1177549881    365.12591 
       5    835.96209    50000 -1222518812 1849842228    365.40199 
```
Notice how Natt goes negative once it exceeds 2^31. In other testing, I've confirmed the rollover happens within 1% of 2^31. That suggests an overflow condition. However, there is no error - the simulation continues with negative Natt. 

I'm running this simulation on a single machine with a serial build, which might lead to having more events handled in one place than in a parallel run, but I don't know what effect that may have. Certainly Sparta has been used for much larger numbers of particles in parallel settings (Gallis et al. 2014 https://www.osti.gov/servlets/purl/1315088). Perhaps this issue did not appear because the timestep was chosen more suitably than the exaggerated value here, reducing Natt per timestep. 

In the case that initially turned up this rollover, I was trying to initialize a volume of gas at near-continuum conditions. In that problem, the gas flows in from one inlet, so a significant amount of time is required for the particles to fill out the volume. I started the run with a timestep that's about 100x longer than the "correct" value given the mean collision time, let it run for 100 steps, then changed back to the correct value and ran a few thousand steps to create many sets of independent, steady-state results to average. 

It's not clear to me whether Natt rolling over has any impact on the simulation results since it's just summing up events across all grid cells to display that total to the user. Is there any limitation internally on the number of collisions or attempted collisions that would render the results unphysical? 

Is there a better way to initialize a simulation like this? Perhaps scale_particles is the way to go.

Run `in.collide` with the changes above.

### LINKED ISSUE #669: surf_react adsorb: when per-sync adsorption overshoots surface capacity, (1-theta) goes negative and ALL surface chemistry dies silently for the rest of the run — is this a bug? (patch included)
We think we've hit a bug and would appreciate your confirmation before we send
a PR. GS adsorption events accumulate in `species_delta` against *stale*
coverage for a whole `Nsync` window, and `update_state_face/surf()` folds them
into the stored state with no capacity check — so when one window's influx
exceeds the face's site capacity, `total_state` exceeds `maxstick` and the
coverage `theta` exceeds 1. Every sticking branch in `react()` then computes
`S_theta = pow((1-theta), n)` with `theta > 1`: a *negative* probability. The
sum of reaction probabilities goes negative, `scatter_prob = 1 - sum_prob`
exceeds 1, and `if (react_prob > random_prob) return 0;` scatters every
subsequent collision — every channel (including ones whose own probability is
positive and physical) is dead for the rest of the run, with no warning and a
clean exit. Are we reading this right?

## Summary
One over-capacity sync window permanently poisons a face/surf: all adsorb
chemistry on it silently stops. Triggerable with small faces, low site
density, long sync windows, or large fnum — per-face `maxstick` can be O(10)
even when the deck looks reasonable.

## Version / reproduction status
- Reproduced on STOCK UPSTREAM `sparta/sparta` master = `5102138` (built
  fresh, unmodified), 1 MPI rank, FACE mode.
- Poison arm (`max_cover 3e4` -> maxstick(zlo) = ceil(3e4*484/1e6) = 15;
  beam ~200 hits/step, nsync 100): 83 reactions in window 1 (i.e. state 83 >>
  capacity 15, theta ~ 5.5 after the fold), then **nsreact == 0 for the
  remaining 1900 steps**; exit 0; no diagnostic anywhere.
- Control arm (identical deck, `max_cover 6.022e18`): chemistry continues the
  whole run (cumulative nsreact 5708).
- The reaction file has two channels: AA adsorption *and* a CI channel whose
  probability (k*theta) is positive and grows with theta — the CI channel is
  killed too, which shows the poisoning acts through the negative sum /
  scatter-everything path, not through any physical saturation.

## Reproducer (attached)
```
spa_mpi -in in.r6_poison    # stats nsreact: >0 in first window, 0 forever after
spa_mpi -in in.r6_control   # identical except max_cover: continues all run
```

## Mechanism (source, verified on 5102138)
- No capacity clamp at the fold: `update_state_face()` /
  `update_state_surf()` apply the deltas with only a lower clamp
  (`species_state = MAX(0, ...)`, surf_react_adsorb.cpp:1406, :1499); nothing
  bounds `total_state` by `maxstick`.
- Negative sticking: in `react()`, every non-Kisliuk sticking branch computes
  `S_theta = pow((1-surf_cover), r->coeff[coeff_val])` (AA :719, DA :740,
  LH1 :771, LH3 :791, CD :811) — negative (or NaN for non-integer exponents)
  once `surf_cover > 1`. (The Kisliuk branches already guard with
  `if (surf_cover < 1)`.) The ER branch's `(maxstick - total_state)` factor
  (:824-825) also goes negative.
- Poisoning: `sum_prob < 0` makes `scatter_prob = 1.0 - sum_prob > 1`
  (:891-892), so `react_prob > random_prob` for every collision (:902) —
  permanent all-scatter with zero feedback to the user.

## Suggested fix (patch attached)
Minimal and conservative — clamp the sticking factors at zero, mirroring the
guard the Kisliuk branches already have:
```cpp
if (surf_cover < 1.0)
  S_theta = pow((1-surf_cover),r->coeff[coeff_val]);   // else stays 0.0
```
in the five branches, plus `MAX(0.0, maxstick - total_state)` in ER. Tested on
5102138: the poison deck recovers (chemistry continues, cumulative 5185; the
CI channel drains the overshoot and adsorption resumes); the control deck's
stats are **bit-identical** to stock (guards dormant whenever theta < 1).
This removes the permanent silent poisoning. The fold itself can still
overshoot capacity within one window (state briefly > maxstick); if you'd
like, a capacity clamp at the fold (with an overflow counter, never silent)
could be a follow-up — the attached one-screen patch is the safe first step.

## Context
- Severity: silent wrong physics — a run that looks healthy (exit 0, particles
  flowing) has zero surface chemistry after some step. Marching-cubes slivers
  and coarse fnum make per-face capacities of O(1-10) easy to hit in implicit
  geometry even at production site densities.

[adsorb_capacity_overshoot.zip](https://github.com/user-attachments/files/31090244/adsorb_capacity_overshoot.zip)

### LINKED ISSUE #670: fix ablate decrement() silently drops the part of a cell's decrement its corners cannot pay (imin == -1 break) — 88% of the requested decrement vanishes untracked on a small stock deck — intended behavior or bug? (patch included)
We think we've found a silent mass-conservation gap and would appreciate your
view before proposing anything bigger. In `FixAblate::decrement()` the greedy
corner-payment loop ends with

```cpp
      if (imin == -1) break;        // fix_ablate.cpp:1042
```

Once every solid corner of the cell has been consumed (or none remains), the
loop breaks with `total > 0.0` and that remainder is never applied anywhere,
never carried forward, and never reported. For ablation driven by a surface
reaction tally this means gas products were already created for solid mass
that is never actually removed — the two sides of the chemistry stop matching,
exactly in the late/burn-through regime. Are we reading this right, or is
dropping the remainder intended?

## Summary
Per-cell decrement demand that exceeds the cell's remaining corner values is
silently discarded. Using only `fix ablate`'s own outputs (global scalar = sum
of unique corner values; vector[1] = sum of requested decrements of the last
epoch), a small stock-style deck shows the total requested decrement exceeding
the total corner mass actually removed by 8.6x, with no warning of any kind.

## Version / reproduction status
- Verified on STOCK UPSTREAM `sparta/sparta` master = `5102138` (built fresh,
  unmodified), 4 MPI ranks; no source changes needed for the measurement.
- Numbers from the attached `in.r5_unpaid` (150x150 grid, the stock
  `binary.101x101` body from examples/ablation, `fix ablate ... random 30`,
  200 epochs):
  - initial corner sum 717,570; final 0
  - total requested (sum of f_ablate[1] over epochs): 6,204,521
  - total actually removed (scalar drop): 717,570
  - silently dropped: 5,486,951 = 88.4% of requested
  - the FIRST epoch is already short (requested 32,110, paid 22,987)
  - late epochs: body fully consumed, requested ~31,000/epoch, paid 0.0
- `grep -i warning` on the log: nothing related.

## Reproducer (attached)
```
spa_mpi -in in.r5_unpaid        # stats print f_ablate (scalar) and f_ablate[1]
# per-epoch conservation check: scalar_prev - scalar_now  vs  f_ablate[1]
```
(The random source makes the demonstration deterministic-in-shape and fast; a
chemistry-driven source produces the same effect whenever a nearly-empty cell
tallies more removal in a sync window than its corners hold, e.g. deep
recession / burn-through.)

## Mechanism (source, verified on 5102138)
- `decrement()` (fix_ablate.cpp:1013-1053): per cell, `total = celldelta[icell]`
  is paid greedily from the smallest positive corner; each corner is used at
  most once (`cdelta[icell][i] == 0.0` test at :1036); when no eligible corner
  remains, `if (imin == -1) break;` (:1042) exits with the unpaid `total`
  discarded.
- Nothing upstream of the surface chemistry checks cell inventory before
  emitting gas products, so "pay-now gas / bill-later solid" diverges by
  exactly the dropped amount.
- A second, related uncounted path: `sync()` (:1076+) applies
  `newvalue = MAX(oldvalue-dsum,0)` for a corner shared by N cells — two cells
  can each claim the same corner's full value in the same epoch and the excess
  is also silently absorbed by the MAX clamp. It is much smaller in practice
  (0.2% of requested in the attached run vs 88.2% via the break).

## Suggested minimal step (patch attached)
We don't presume the right physics answer (carry-forward? neighbor spill?
error?), but silent dropping seems clearly unwanted. The attached minimal
patch just makes the loss visible:
- counts the unpaid remainder per epoch (`decrement()` only),
- prints a one-time WARNING the first time it happens,
- exposes the cumulative total as a 3rd element of the fix's global vector.
Tested on 5102138: warning fires on the attached deck; scalar-drop +
unpaid-tally accounts for 99.79% of the requested total (the 0.21% residual is
the shared-corner MAX clamp above, which the patch deliberately leaves alone).
The stock example decks (no over-demand) never trigger the warning.

## Context
- Severity: correctness of coupled mass budgets in late-stage ablation;
  invisible in short/benign runs, systematic in burn-through regimes.
- If you'd prefer a conserving treatment (carry the deficit forward per cell
  and pay it from whatever solid remains at later syncs, spilling to a
  face-adjacent solid cell at burn-through), we're happy to draft that as a
  follow-up PR — the counting patch is deliberately the smallest first step.

[fixablate_imin_break.zip](https://github.com/user-attachments/files/31090345/fixablate_imin_break.zip)

stanmoore1:
## Purpose

Many collected small changes and bugfixes, see commit messages for description
Fixes #669
Fixes #670
Fixes #492

## Author(s)

Stan Moore (SNL), Claude Code (Opus/Fable 5), Upendra Yadav (University of California, Irvine)

## Backward Compatibility

Yes


stanmoore1:
Regression tests passed.

PR #673, opened 2026-08-19
Port LAMMPS dump image rendering enhancements and transparency

### LINKED ISSUE #671: Add transparency to entities in image dumps
## Summary

It would be nice to have the option to render particles, surfaces, and other objects as partially transparent when dumping images. 

## Type of Issue

Suggestion for an Enhancement

While dump_image/dump_modify support custom colors for any visible object, neither presently permits transparency. This can make it difficult to see the interiors of dense regions or finely-meshed surfaces. I know that the existing rendering procedures have been designed for speed when dealing with large numbers of particles and I suppose that adding transparency might slow down the rendering. However, such a trade-off is evidently acceptable to some users; the ssao option, for example, takes a long time but still finds use.

stanmoore1:
## Purpose

Ports two sets of `dump image` rendering features from LAMMPS to SPARTA, then follows up with cleanup of dead code the ports surfaced.

Closes #671 

**New `dump image` keywords** (from [lammps/lammps#5105](https://github.com/lammps/lammps/pull/5105)):

- `depthcue` — fade distant objects toward a fog color, with selectable color and start position
- `defocus` — blur distant objects with a disk-shaped aperture kernel
- `outline` — draw outlines at depth discontinuities, for a flat illustration look

**New `dump_modify` keywords:**

- `gamma` — tonal adjustment of rendered objects
- `specular` — `none`/`wide`/`narrow`/`tight` presets controlling highlight width independently of `shiny`
- `metal`, `metalfinish` — metallic surface shading, satin through mirror
- `ssaosamples` — trade SSAO quality against speed
- `ptrans`, `gtrans`, `strans`, `glinetrans`, `slinetrans`, `boxtrans`, `subboxtrans`, `axestrans` — per-object transparency

**SSAO is now deterministic.** The per-rank RNG is replaced with interleaved gradient noise keyed on pixel position and the `ssao` seed, so images no longer depend on the number of MPI ranks and are exactly reproducible for an unchanged scene. This removes flickering shading in movies. It is also the one rendering change that is intentionally *not* bit-compatible with master — see below.

**Also fixed:** the doc page told users to change per-type particle diameters with `dump_modify adiam`, which is the LAMMPS name and does not exist in SPARTA. Following it produced `ERROR: Illegal dump_modify command`. It is `pdiam`.

## Author(s)

Stan Moore (SNL) and Claude Code

The original LAMMPS implementations being ported are by Axel Kohlmeyer (Temple University) — the rendering enhancements in lammps/lammps#5105, and the screen-door transparency in lammps/lammps commits `6561ebbf6a`, `f3137c521d` and `32a7db5431`.

## Backward Compatibility

**Two intentional changes in observable behavior.**

1. **SSAO images differ from master.** This is the point of the determinism change: shading now comes from a deterministic noise pattern rather than a per-rank RNG. Any scene using `ssao yes` renders differently than before (and identically across rank counts, which it did not before). Everything not using `ssao` is bit-for-bit unchanged.

2. **The non-functional `persp` keyword is removed.** Its parser called `error->all()` as its first statement, so using it always aborted the run — the renderer is a parallel-ray rasterizer with no perspective code path. A script containing `persp` previously aborted with "Dump image persp option is not yet supported" and now aborts with "Illegal dump image command". Both abort, so nothing that previously worked stops working; only the wording of an existing failure changes. This matches LAMMPS, which carried the identical non-functional keyword with word-for-word identical documentation and removed it in `405c1b98c5` (Sep 2020), then removed the vestigial `Image::persp` member in `5ca36afc4a` (Jan 2026).

Also removed, with no observable effect: `Image::gridcolor` (declared, never initialized, never used) and `Image::element2color()` / `Image::element2diam()` (public but with no callers anywhere in the tree). The element functions are only useful with a `dump_modify element` keyword, which SPARTA does not have — the diameters are AtomEye per-element radii in Ångströms used directly as a length, meaningless when SPARTA's unit styles are `si` and `cgs`, and most SPARTA species IDs are molecules and ions (`N2`, `O2`, `NO`, `CO2`, `O2+`, `e`) absent from the 109-element table. Recoverable from git history if ever wanted.

Everything else is additive, and all new options default to off / no-op.

## Implementation Notes

**Transparency is screen-door, not alpha blending.** A 16×16 Bayer threshold matrix decides per screen pixel whether a fragment is drawn. Skipped pixels are never passed to `draw_pixel()`, so every written pixel stays fully opaque and `draw_pixel()`, the depth buffer and the nearest-depth-wins compositing in `merge()` are untouched. Because the pattern is anchored to screen coordinates it is identical on every rank, so rank-independence is preserved by construction. Recommended values are 0.25/0.5/0.75 with `fsaa yes`, whose 2×2 downsample turns the stencil into something that reads as a blend.

**Three places this deliberately differs from LAMMPS:**

1. **No `diameter <= 0.0` entry guard.** LAMMPS has one; transliterating it would silently delete the `gridx`/`gridy`/`gridz` feature, since those call `draw_brick()` with a zero extent on purpose. Zero diameters are reachable elsewhere too — `sdiamvalue` is unvalidated and `glinediam`/`slinediam` only reject negatives.
2. **The back-face cull in `draw_triangle()` is kept.** A back-face normal points away from the camera and would feed wrong values into the diffuse and grazing terms of `draw_pixel()`. What users want to see through a transparent surf is the particles and grid cells inside it, which are separate draws that win the depth test in the dither holes.
3. **`ptrans` rather than LAMMPS' `atrans`**, matching SPARTA's neighbouring `pcolor`/`pdiam`. Four of the eight transparency keywords have the same name as in LAMMPS; the other four cover objects only SPARTA draws.

**One bug fixed relative to upstream.** `compute_depthcue()` and `compute_defocus()` guarded a degenerate depth range against an absolute `EPSILON = 1.0e-6`. That is fine in LAMMPS, whose length unit is Ångströms, but SPARTA uses SI metres, so a sub-micron box made both effects silently do nothing. The guards now scale with the box size, as `compute_outline()` already did. Worth reporting upstream, since LAMMPS supports `units si` too.

**Known limitation, documented not fixed:** transparency should not be combined with `outline`. A dither hole is indistinguishable from background in the depth buffer, so a transparent object is outlined pixel by pixel and floods with outline color. LAMMPS has the same defect; the fix needs a per-pixel hole mask surviving `merge()`, which is larger than the whole transparency port. The milder interactions with `ssao` and `depthcue`, and the beneficial one with `defocus`, are documented too.

### How correctness was verified

- **No regression on existing features.** With `ssao` off, rendered `.ppm` output is byte-identical to a binary built from `master` on a scene exercising every pre-existing image option (grid, grid cutting planes, `gline`, surf, `sline`, box, subbox, axes, `shiny`, `fsaa`, `backcolor2`, `lights`), and for all 16 images of the shipped `examples/custom` case. With `ssao` on the output differs, by design.
- **New keywords are no-ops at their defaults**, verified byte-for-byte across the branch's own history.
- **Rank independence.** A deterministic scene with all eight transparency keywords active renders byte-identical at 1, 3 and 4 ranks. The same scene under the pre-port code differs between 1 and 4 ranks, confirming the SSAO change is real rather than vacuous. `-np 3` is included to exercise the non-power-of-two branch of the merge tree and the `MPI_Gatherv` path.
- **Dither correctness.** Each of the four rasterizers hits the predicted drawn fraction to within 0.0006 at opacities 0.25/0.5/0.75, and opacity 0.0 renders byte-identical to the object being switched off. The Bayer table was validated as an exact permutation of (k+0.5)/256 and against an independently computed Bayer recursion.
- **Depth semantics intact.** An opaque object in front of a transparent one is never punctured.
- **The epsilon fix is falsifiable.** On a 100 nm box, `depthcue` produced no fading before the fix (mean luminance of drawn pixels 67.4) and fades correctly after (39.5); `defocus` likewise.
- **Physics untouched.** Five shipped examples (`custom.step.set`, `custom.cube.set`, `collide`, `sphere`, `circle`) produce identical stats output against a `master` build, timing excluded.
- Clean from-scratch builds of `serial` and `mpi`, no new compiler diagnostics under `-Wall -Wextra` against a `master` baseline.

Rebased onto current master (`6f7d62d7`) and re-verified there, including that master's recent `dump_image.cpp` fixes (binary-mode file open, the `up` vector y-component fix, the dump movie filename fix) are preserved.

## Post Submission Checklist


stanmoore1:
Sample images:

dump_image_metal

dump_image_postproc

dump_image_shading

dump_image_transparency

PR #618, opened 2026-05-27
Large AI assisted bug fix commit

aborner1:
## Purpose

Went through the code with various LLM agents and debugged. Sent Stan email with list of bugs. This commit has all the fixes.

## Author(s)

Arnaud Borner (NASA Ames) + Claude Code

## Backward Compatibility

Should be backward compatible as it's bug fixes

## Implementation Notes

Fixing bugs only

## Post Submission Checklist

- [x] The feature or features in this pull request is complete
- [N/A] Suitable new documentation files and/or updates to the existing docs are included
- [N/A] One or more example input decks are included
- [x] The source code follows the SPARTA formatting guidelines

## Further Information, Files, and Links

ablation.3d test fails due to changes. exp2imp test was already failing due to erroring out: "Calculated solid fraction above one or negative”. Many tests need to be reblessed with new gold standards:
       53 - mpi.in.beam.surf.gs_ps.mpi_1 (Failed)
       54 - mpi.in.beam.surf.ps.mpi_1 (Failed)
       62 - mpi.in.beam.surf.gs_ps.mpi_4 (Failed)
       63 - mpi.in.beam.surf.ps.mpi_4 (Failed)
       97 - mpi.in.collideInterspecies.mpi_1 (Failed)
       98 - mpi.in.collideInterspecies.mpi_4 (Failed)
      111 - mpi.in.thermostat.mpi_1 (Failed)
      112 - mpi.in.thermostat_ave.mpi_1 (Failed)
      113 - mpi.in.thermostat.mpi_4 (Failed)
      114 - mpi.in.thermostat_ave.mpi_4 (Failed)
      115 - mpi.in.bfield.mpi_1 (Failed)
      116 - mpi.in.bfield.grid.mpi_1 (Failed)
      117 - mpi.in.bfield.mpi_4 (Failed)
      118 - mpi.in.bfield.grid.mpi_4 (Failed)
      119 - mpi.in.circle.adjust.mpi_1 (Failed)
      120 - mpi.in.circle.constant.mpi_1 (Failed)
      121 - mpi.in.sphere.adjust.mpi_1 (Failed)
      122 - mpi.in.sphere.constant.mpi_1 (Failed)
      123 - mpi.in.circle.adjust.mpi_4 (Failed)
      124 - mpi.in.circle.constant.mpi_4 (Failed)
      125 - mpi.in.sphere.adjust.mpi_4 (Failed)
      126 - mpi.in.sphere.constant.mpi_4 (Failed)
[...]

PR #675, opened 2026-08-20
Consolidate the verified static-analysis bug fixes from #618

stanmoore1:
## Purpose

Consolidates #618 changes into a single set of fixes for genuine defects only.
Closes #618

Highlights of applied fixes:

- **Wrong results / logic**: tally keyword and enum typos (`type2`→TYPE1, `vy2`/`vz2` shift, duplicated ID2POST test), compute count `imix`/`igroup` mixup, `tvib` per-mode `groupspecies` indexing OOB, `property/surf` 3D pack OOB, `reaction2col` wiped inside the strtok loop, `rotc2` not set symmetrically, NaN comparison accepting zero-relative-velocity collisions when `vremax` is zero, `react_qk`/`react_tce_qk` using `react_prob` as rejection-loop scratch, `react_tce` unreachable warnings and missing `break`, `react_bird` uninitialized tally pointers, `adapt_grid` coarsen cell index, marching-cubes `int` corner-value truncation, specular `wrapper()` ignoring `noslip`, `emit/face/file` half-circle azimuth, `tally_setup` dangling active-list pointers.
- **KOKKOS**: RNG pool state freed then reused in the collide attempt loop, `return` vs `continue` in per-cell particle loops (eflux/thermal), lambda/grid missing `else` writing an unallocated array (CPU too), sonine dead OOB device read, `tallyinfo` OOB read when `nsurf` is zero, `ave/histo(/weight)` inverted realloc conditions and stray `printf`, grid-check OOB host read, `compute_surf` ECHEM conditional column advance (CPU too), `bin_particles` reading an uninitialized `Region` pointer when no region is defined.
- **Crashes / hangs / NaN**: mover box-exit fraction 0/0 with clamp, stuck-particle epsilon (tiny positive advances hung forever), `axi_remap` on-axis division, axisymmetric quadratic catastrophic cancellation (Vieta) plus the degenerate linear case, `temp/rescale` cold-cell division, ablate `Ninterface` zero division, subsonic emit zero-division, `fix surf/temp` uninitialized units and stale compute/fix pointers, universe-variable lock-file fopen/fscanf, `read_isurf` unchecked binary fopen.
- **Hardening**: `snprintf` for unbounded `%s` filename/name messages across the code base (input, write_*, read_*, sparta, output restart names with BIGINT timesteps, domain/modify/surf suffix styles, mixture, fix_print, surf_react_prob, PYTHON), a truncation-safe rewrite of the `stats.cpp` line/format building (user `stats_modify format` strings could overflow the 32-byte per-field rows and the line buffer), `str[32]` cell-ID buffers grown and bounded, `BIGINT_FORMAT` in `fix_halt`, and memory leaks on recoverable error paths (SPARTA errors throw, so library callers can recover and would otherwise accumulate the leaks).
- **KOKKOS FFT convention**: `fft_2d_1d_only_kokkos` and `fft_3d_1d_only_kokkos` ran the forward plans when flag was -1 and applied normalization when flag was 1 — the opposite of the host `fft2d.cpp`/`fft3d.cpp` and inconsistent with their own cuFFT/hipFFT branches. Both now match the CPU convention (timing-only routines; no simulation output change).

Intentionally **not** applied, each with a verified rationale: the one-line fft2d scaling flip from upstream #618 (a regression given the inverted branch labels in that routine — superseded by the full convention fix above), the `surf_collide_piston` zero-`uprime` guard from #618 (provably unreachable: the sign of `xprime - xwall` always triggers the existing inclusive deletion checks in that case), inert integer casts (`int` times `sizeof` already promotes to 64-bit, and `Memory::create` takes an `int` count), unreachable defensive guards (dt/grid `vrm_max`; the collide volume guard — all callers hard-error first; the `vr2` scatter guard, unreachable once the `vremax` fix is in; the `ecc` guards, implied by upstream `e_excess` checks; adsorb zero-velocity), RNG seed and idiom changes that shift reproducibility baselines (`-log(1.0-drand())`, pool seeds), and cosmetic `snprintf` conversions of provably bounded formats.

Three applied fixes change physical output and are flagged for domain sign-off: the `react_tce` missing `break` (chem-rate tallies), the axisymmetric Vieta/degenerate-case geometry change, and the `emit/face/file` `MY_2PI` azimuth.

## Author(s)

Arnaud Borner (NASA Ames), Stan Moore (SNL), and Claude Code

## Backward Compatibility

No input-syntax changes. Numerical output changes only where the previous behavior was defective: chem-rate tallies with `computeChemRates` (missing `break`), tangential velocities from `fix emit/face/file` (azimuth was half-circle), `compute surf` column alignment when `echem` is not the last value, axisymmetric surface intersections at roundoff level, and states that previously produced NaN or hangs. `fix surf/temp` now errors on unit styles other than si/cgs instead of using uninitialized values. Over-long user filenames and format strings now produce truncated messages instead of stack corruption.

## Implementation Notes

- Every candidate fix was checked against current `master` first; entries already fixed by intervening work (comm double-alloc, timer, CLL/impulsive destructors, `idsource` shadow, `memory_usage`, `size_restart` redesign, several memset widenings, and about twenty more) were skipped rather than re-applied.
- CPU/KOKKOS parity was preserved: each fix was applied to both variants where both were affected, and KOKKOS-only defensive guards with no CPU twin were not introduced.
- The full per-bug decision table (bug number mapped to applied / already fixed upstream / rejected, with rationale) and the coverage accounting of sparta/sparta#618 are provided separately rather than in-repo.

## Verification

- **Line-by-line re-audit of all 111 reported bugs.** A scripted pass ran 74 automated pattern checks over every applied fix site on the final branch, asserting the corrected code is present and the buggy pattern is gone; all pass. The checks cover both the CPU and KOKKOS variant of each fix. This pass is what caught the one gap (the `fix_ablate` suffix leak, masked because the neighboring `idsource` bug was already fixed on master) — it was then fixed and re-checked.
- **Every rejection re-proved from current code, not taken on faith from the audits**: all four `attempt_collision` callers hard-error on zero cell volume before the call; a zero-velocity or tangential particle cannot register the surface hit that would reach the adsorb divisions; the piston zero-`uprime` division is unreachable because the sign of `xprime - xwall` then always satisfies the inclusive deletion checks; the dt/grid guard sits below an upstream temperature check that already excludes it; the `ecc` guards are implied by the `e_excess` early-continue; the fft2d one-line flip was traced through all five backend branches of the timing routine to confirm it would pair normalization with the wrong transform.
- **Fix-variant correctness was checked where the two source branches disagreed**: the rejection-sampling fix uses the separate scratch variable (keeping `react_prob` as scratch would leave the sampled value in the accumulator that the downstream reaction test reads — the alternative fix circulating in the bug list is itself buggy), and the `stats.cpp` port replaces the upstream version, whose offset arithmetic advances by the untruncated `snprintf` return value and could walk past the buffer.
- **Deliberately untouched code confirmed untouched**: the `fix_ablate` `epsilon_adjust()` threshold logic (the known trap flagged in the bug list) is byte-identical to master.
- **Builds**: `make serial` compiles and links `spa_serial` cleanly after each commit; KOKKOS-only and package files (KOKKOS, PYTHON, FFT) received minimal targeted edits that mirror their verified CPU twins.
- Not yet run: the example-deck regression suite and a KOKKOS build; the three flagged output-changing fixes (react_tce `break`, axisymmetric geometry, emit azimuth) are expected to shift chem-rate tallies, axisymmetric trajectories at roundoff level, and emitted tangential velocity distributions respectively.

## Post Submission Checklist

### DIFF
--- a/src/KOKKOS/collide_vss_kokkos.cpp
+++ b/src/KOKKOS/collide_vss_kokkos.cpp
@@ -982,7 +982,6 @@ void CollideVSSKokkos::operator()(TagCollideCollisionsOne< NEARCP, GASTALLY, ATO
       else
         reduce.nreact_one++;
     } else {
-      rand_pool.free_state(rand_gen);
       continue;
     }
 
@@ -2590,7 +2589,6 @@ void CollideVSSKokkos::operator()(TagCollideCollisionsOneAmbipolar< GASTALLY, AT
       else
         reduce.nreact_one++;
     } else {
-      rand_pool.free_state(rand_gen);
       continue;
     }
 
--- a/src/KOKKOS/compute_eflux_grid_kokkos.cpp
+++ b/src/KOKKOS/compute_eflux_grid_kokkos.cpp
@@ -241,7 +241,7 @@ void ComputeEFluxGridKokkos::operator()(TagComputeEFluxGrid_compute_per_grid, co
 
     const int ispecies = d_particles[i].ispecies;
     const int igroup = d_s2g(imix,ispecies);
-    if (igroup < 0) return;
+    if (igroup < 0) continue;
 
     const double mass = d_species[ispecies].mass;
     double *v = d_particles[i].v;
--- a/src/KOKKOS/compute_lambda_grid_kokkos.cpp
+++ b/src/KOKKOS/compute_lambda_grid_kokkos.cpp
@@ -326,12 +326,12 @@ void ComputeLambdaGridKokkos::compute_per_grid_kokkos()
 
       if (l_knyflag) {
         if (l_noutputs == 1) l_vector_grid[i] = lambda / sizey;
-        l_array_grid(i,l_output_order[KNY]) = lambda / sizey;
+        else l_array_grid(i,l_output_order[KNY]) = lambda / sizey;
       }
 
       if (l_knzflag) {
         if (l_noutputs == 1) l_vector_grid[i] = lambda / sizez;
-        l_array_grid(i,l_output_order[KNZ]) = lambda / sizez;
+        else l_array_grid(i,l_output_order[KNZ]) = lambda / sizez;
       }
     });
   }
--- a/src/KOKKOS/compute_sonine_grid_kokkos.cpp
+++ b/src/KOKKOS/compute_sonine_grid_kokkos.cpp
@@ -225,9 +225,6 @@ void ComputeSonineGridKokkos::operator()(TagComputeSonineGrid_compute_vcom, cons
 
 KOKKOS_INLINE_FUNCTION
 void ComputeSonineGridKokkos::operator()(TagComputeSonineGrid_normalize_vcom, const int &icell) const {
-  const int ispecies = d_particles[icell].ispecies;
-  const int igroup = d_s2g(imix,ispecies);
-
   double norm;
   for (int j=0; j<ngroup; j++) {
     norm = d_vcom(icell,j,3);
--- a/src/KOKKOS/compute_surf_kokkos.cpp
+++ b/src/KOKKOS/compute_surf_kokkos.cpp
@@ -226,7 +226,7 @@ int ComputeSurfKokkos::tallyinfo(surfint *&ptr)
 
   while (1) {
     while (istart < nsurf && h_surf2tally[istart] != -1) istart++;
-    while (h_surf2tally[iend] == -1 && iend > 0) iend--;
+    while (iend > 0 && h_surf2tally[iend] == -1) iend--;
     if (istart >= iend) {
       ntally = istart;
       break;
--- a/src/KOKKOS/compute_surf_kokkos.h
+++ b/src/KOKKOS/compute_surf_kokkos.h
@@ -409,8 +409,9 @@ void surf_tally_kk(double /*dtremain*/, int isurf, int icell, int reaction,
         double r_coeff = 0.0;
         if (sr_type == 1)
           r_coeff = sr_kk_prob_copy[m].obj.d_coeffs(reaction-1,1);
-        a_array_surf_tally(itally,k++) += weight * r_coeff * fluxscale;
+        a_array_surf_tally(itally,k) += weight * r_coeff * fluxscale;
       }
+      k++;
       break;
     case ETOT:
       if (iorig) vsqpre = origmass * MathExtraKokkos::lensq3(vorig);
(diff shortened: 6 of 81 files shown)


stanmoore1:
@aborner1 Here is Fable's reply:

Yes — your instinct is exactly right. Here's the full picture.

## What the revert actually does

The two `if` statements sit between the `break;` that closes the RECOMBINATION case and the `default:` label, *inside* the switch. In C++, statements in a switch body are only reachable via a case label or fallthrough, and every path above them ends in `break` or `continue`. So they are dead code — **putting them back doesn't "restore" master behavior so much as re-mute the warning entirely; it has never fired in any released version.** The revert traded a broken diagnostic for no diagnostic, purely to keep CI/reference logs stable.

## Is the check itself meaningful? (correcting my own earlier claim)

I told you earlier the check "misfires because the cumulative roulette probability legitimately exceeds 1" — **that was wrong, and I want to correct it.** `react_prob` accumulates P_i over the candidate reactions, and `random_prob ∈ (0,1)` is drawn once. The scheme is only valid when ΣP_i ≤ 1 (the remainder being "no reaction"). If the cumulative crosses 1, a reaction becomes *guaranteed* and all probability mass past 1 is silently clipped — i.e., the chemistry is under-resolved and rates are biased. So `react_prob > 1` is precisely the validity-violation tripwire the original author intended, and it's real: my serial `in.chem` run tripped it 5 times in 1000 steps. The `< 0` branch similarly catches bad reaction-file coefficients. (It's a partial detector — once a reaction is selected the routine returns, so clipping past the selection point goes unseen — but it catches the egregious cases.)

## Why the naive fix (just moving it after the switch) is still wrong

`Error::warning` prints **unconditionally on every calling rank** — there's no `comm->me == 0` gate and no once-only logic. `attempt()` runs per candidate collision on every rank, so a production deck that clips would emit one line *per clipped collision per rank*: potentially millions of lines, plus it breaks every chem reference log. That's why simply relocating it failed CI.

## Better solution: warn once per run

The standard pattern (LAMMPS-style flagged warning):

- add an `int clip_warn_flag;` member to `ReactTCE`, reset in `init()`;
- after the switch: on first `react_prob > 1.0` (or `< 0.0`) set the flag and emit **one** warning with an actionable message, e.g. *"TCE reaction probability exceeded 1 — chemistry may be under-resolved; consider reducing fnum or the timestep"*;
- cost in the hot loop is one predictable branch; the KOKKOS `attempt_kk` stays as-is (device code can't warn — acceptable, with plenty of precedent).

